PR #4, opened 2026-06-07
Fix AI error surfacing, onboarding reminder, Timestamp Detective project_paths wiring, and dynamic git commit ingestion window

### LINKED ISSUE #3: Fix AI error surfacing and Timestamp Detective project_paths wiring
Bug 1: AI silently fails with no feedback
The _send_llm_request function swallows all exceptions and returns None: ai.py:49-63

When installed fresh on someone else's Mac, load_config() returns defaults where active_provider is "disabled": config.py:113-116

So AI is off by default. The user needs to run:

termstory config set active_provider groq  
termstory config set providers.groq.api_key <their_key>
But even if they do configure it and something goes wrong (wrong key, network issue, wrong URL), the bare except Exception: return None means the TUI just shows nothing — no error, no indication of what failed. There's no way to diagnose it.

Bug 2: Timestamp Detective always gets project_paths=[]
This is the core problem with old logs not working. In parser.py, the TimestampDetective is always initialized with an empty project_paths: parser.py:187-191

The comment says "populated from cli.py via parse_all_histories when available" — but looking at run_ingestion() in cli.py, it calls parse_all_histories(history_files, db=db), and parse_all_histories calls parse_zsh_history(path, existing_lookup) — no project paths are ever passed down. The project_paths argument to TimestampDetective is hardcoded to []. parser.py:418-437

This matters because the git commit detector builds its repo search list like this: timestamp_detective.py:416-424

With project_paths=[], the only repo it can search is the one derived from the virtual CWD at the time of each git commit command. If the virtual CWD doesn't resolve to a valid git root (e.g., the path doesn't exist on this machine, or the history doesn't have cd commands leading there), the git commit detector finds nothing and the command gets a synthetic timestamp.

On someone else's Mac, their repos are likely in different paths than what the virtual CWD tracker resolves to, so the Detective gets zero git anchors and falls back to the mtime step-back — which collapses everything to "today."

Secondary issue: commit ingestion is capped at 90 days
Even after the Detective runs, run_ingestion() only fetches commits for session enrichment from the last 90 days: cli.py:84-90

This doesn't affect the Detective (which uses git log --all with no time filter), but it means sessions older than 90 days won't have commit data attached to them even if the Detective correctly placed them.

What needs fixing
AI: Add error surfacing in _send_llm_request — at minimum log to stderr or set a flag the TUI can display. Also add a first-run prompt or termstory config list reminder when AI is disabled.

project_paths wiring: parse_all_histories needs to accept and pass project_paths down to parse_zsh_history, which passes it to TimestampDetective. The run_ingestion() in cli.py should discover project paths (e.g., by scanning ~/Projects, ~/src, or known git repos) and pass them in.

Fix AI error surfacing and Timestamp Detective project_paths wiring
There are two bugs to fix in bitflicker64/Termstory:

Bug 1: AI silently fails with no user feedback
In termstory/ai.py, _send_llm_request (line 49-63) catches all exceptions and returns None silently. When AI is misconfigured or the network fails, the TUI shows nothing.

Fix: Capture the exception message and store it somewhere accessible (e.g., return a tuple (result, error_str) or write to a module-level _last_ai_error variable). The TUI or CLI should surface this error to the user when AI returns None and a provider is configured (i.e., provider != "disabled").

Also, in termstory/config.py, the default active_provider is "disabled" (line 115). When a user installs fresh and runs termstory ui, there's no indication that AI is off. Consider printing a one-time hint (gated by has_seen_onboarding) telling the user to run termstory config set active_provider groq and set their API key.

Bug 2: Timestamp Detective always receives project_paths=[]
In termstory/parser.py (line 187-190), TimestampDetective is always initialized with project_paths=[]. The comment says it should be populated from cli.py, but the wiring is missing.

Fix the call chain:

termstory/parser.py: Change parse_zsh_history(filepath, existing_lookup) to accept an optional project_paths: Optional[List[str]] = None parameter and pass it to TimestampDetective(search_root=..., project_paths=project_paths).

termstory/parser.py: Change parse_all_histories(filepaths, db) to accept an optional project_paths: Optional[List[str]] = None parameter and pass it through to parse_zsh_history and parse_bash_history.

termstory/cli.py: In run_ingestion(db), after detect_projects(sessions) returns the project list, collect the project paths and pass them to parse_all_histories. Since detect_projects runs after parse_all_histories, you need to do a two-pass approach: first parse without project paths to get sessions/projects, then re-parse with project paths if there were legacy commands. Alternatively, discover project paths independently by scanning common directories (~/Projects, ~/src, ~/Developer, ~/Code, ~/Work) for git repos before calling parse_all_histories, and pass those discovered paths in.

The simplest fix for the second approach: in run_ingestion, before calling parse_all_histories, scan common project root directories for .git folders (up to 2 levels deep) and collect those paths. Pass them as project_paths to parse_all_histories. This gives the Timestamp Detective real repos to search even when the virtual CWD tracker can't resolve paths.

Example discovery logic to add to run_ingestion in termstory/cli.py:

import glob  
project_paths = []  
for root_dir in ["~/Projects", "~/src", "~/Developer", "~/Code", "~/Work", "~"]:  
    expanded = os.path.expanduser(root_dir)  
    if os.path.isdir(expanded):  
        for git_dir in glob.glob(os.path.join(expanded, "*", ".git")):  
            project_paths.append(os.path.dirname(git_dir))  
        for git_dir in glob.glob(os.path.join(expanded, "*", "*", ".git")):  
            project_paths.append(os.path.dirname(git_dir))
Pass project_paths to parse_all_histories(history_files, db=db, project_paths=project_paths).

Make sure parse_bash_history also accepts and ignores (or uses) the project_paths parameter for API consistency, even if the Timestamp Detective is only used in the zsh parser currently.

bitflicker64:
close #3


bitflicker64:
@copilot review

PR #13, opened 2026-06-16
fix: replace monkeypatch of CliRunner.mix_stderr with CliRunner(mix_stderr=False)

bitflicker64:


bitflicker64:
@greptileai review

PR #114, opened 2026-06-23
Fix config loading error handling

### LINKED ISSUE #110: [Easy] config.py: `except Exception: config = {}` silently discards JSON …
## 📍 Location

| Field | Value |
|-------|-------|
| **File** | `termstory/config.py` |
| **Line** | 172 |
| **Type** | `TODO` |
| **Category** | `bug` |
| **Difficulty** | Easy |

---

## 🔍 Analysis

Let me look at the relevant code in `termstory/config.py` around line 172.

Let me get more context around the file to understand the full picture, including imports and any other exception handling patterns.

I have enough context to answer the question. Let me compile the answer now.

---

## 1. Context

```python
# 1. Read existing config file
config = {}
if os.path.exists(config_path):
    try:
        with open(config_path, "r", encoding="utf-8") as f:
            config = json.load(f)
    except Exception:       # ← line 172: bare except, no message, no re-raise
        config = {}
if not isinstance(config, dict):
    config = {}
```

The file has no `import logging` and no `print()` calls anywhere — there is zero error-reporting infrastructure in `config.py`.

---

## 2. Root Cause

Three compounding problems:

| # | Problem |
|---|---------|
| 1 | **Bare `except Exception`** catches everything — `json.JSONDecodeError`, `UnicodeDecodeError`, `PermissionError`, `OSError` — all silently. |
| 2 | **No user-visible message.** The caller gets back a silently-defaulted config as if the file simply didn't exist. The user has no idea their config file is corrupt or unreadable. |
| 3 | **Data loss risk.** Later in the function the merged config is written back to disk. If the file was corrupt and the error was swallowed, the corrupt file gets silently overwritten with defaults, destroying whatever was recoverable. |

---

## 3. Suggested Fix

**File:** `termstory/config.py`

**Step 1 — Add `sys` to imports** (already has `os`, `json`):
```python
import sys
```

**Step 2 — Replace the bare `except` with specific, noisy handlers** (lines 169–173):
```python
try:
    with open(config_path, "r", encoding="utf-8") as f:
        config = json.load(f)
except json.JSONDecodeError as e:
    print(
        f"Warning: config file '{config_path}' contains invalid JSON "
        f"and will be ignored ({e}). Fix or delete it to suppress this warning.",
        file=sys.stderr,
    )
    config = {}
except OSError as e:
    print(
        f"Warning: could not read config file '{config_path}': {e}",
        file=sys.stderr,
    )
    config = {}
```

- `json.JSONDecodeError` (a subclass of `ValueError`) covers corrupt/truncated JSON.
- `OSError` covers permission errors, broken symlinks, etc.
- Printing to `stderr` means the warning appears in the terminal without polluting stdout (which may carry structured output).

**Step 3 (optional but recommended) — Guard the write-back** so a corrupt file isn't silently overwritten with defaults. Find where the merged config is saved and only write if the load actually succeeded (or prompt the user first).

---

## 4. Effort

**S** — The core fix (Steps 1 & 2) is a handful of lines. The optional write-back guard depends on how the save path is structured but is still small.

---

## 5. Prerequisites

- **`json.JSONDecodeError`** — Python's built-in exception raised by `json.load()` when the file is not valid JSON. It is a subclass of `ValueError`, available since Python 3.5.
- **`sys.stderr`** — The standard error stream; writing warnings here is the Unix convention so they don't mix with normal program output.
- **XDG Base Directory spec** — The config file lives under `~/.config/termstory/` (or `~/.termstory/` for legacy installs), as set up by `get_app_dir()`. A newcomer should know where to find the file to reproduce the bug manually (e.g., by writing `{bad json` into it).

---

Urvity03:
## Summary

This PR improves error handling when loading the configuration file.

### Changes

* Replaced the broad `except Exception` block with specific exception handling:

  * `json.JSONDecodeError`
  * `OSError`
* Added warning messages to `stderr` when the config file is invalid or cannot be read.
* Added `sys` import for `stderr` output.

### Benefits

* Prevents silent failures when configuration files contain invalid JSON.
* Makes configuration issues visible to users.
* Improves debugging and user experience.
- close #110

### DIFF
--- a/termstory/config.py
+++ b/termstory/config.py
@@ -171,7 +171,13 @@ def load_config() -> dict:
             with open(config_path, "r", encoding="utf-8") as f:
                 config = json.load(f)
 
-        except (json.JSONDecodeError, UnicodeDecodeError) as e:
+        except (
+            json.JSONDecodeError,
+            UnicodeDecodeError,
+            OSError,
+            ValueError,
+            RecursionError,
+        ) as e:
             print(
                 f"Warning: config file '{config_path}' contains invalid data and will be ignored ({e}).",
                 file=sys.stderr,
--- a/termstory/config.py
+++ b/termstory/config.py
@@ -1,5 +1,6 @@
 import os
 import json
+import sys
 from typing import List, Any
 
 def get_app_dir(dir_type: str = "data") -> str:
@@ -163,14 +164,23 @@ def load_config() -> dict:
         "max_history_age": 5
     }
     
-    # 1. Read existing config file
-    config = {}
-    if os.path.exists(config_path):
-        try:
-            with open(config_path, "r", encoding="utf-8") as f:
-                config = json.load(f)
-        except Exception:
-            config = {}
+    #
+    try:
+        with open(config_path, "r", encoding="utf-8") as f:
+            config = json.load(f)
+    except json.JSONDecodeError as e:
+        print(
+            f"Warning: config file '{config_path}' contains invalid JSON and will be ignored ({e}).",
+            file=sys.stderr,
+        )
+        config = {}
+    except OSError as e:
+        print(
+            f"Warning: could not read config file '{config_path}': {e}",
+            file=sys.stderr,
+        )
+        config = {}
+
     if not isinstance(config, dict):
         config = {}
             
@@ -229,6 +239,7 @@ def load_config() -> dict:
         save_config(config)
         
     return config
+
 
 def save_config(config: dict) -> None:
     """Save configuration dictionary to disk"""
--- a/termstory/config.py
+++ b/termstory/config.py
@@ -164,22 +164,26 @@ def load_config() -> dict:
         "max_history_age": 5
     }
     
-    #
-    try:
-        with open(config_path, "r", encoding="utf-8") as f:
-            config = json.load(f)
-    except json.JSONDecodeError as e:
-        print(
-            f"Warning: config file '{config_path}' contains invalid JSON and will be ignored ({e}).",
-            file=sys.stderr,
-        )
-        config = {}
-    except OSError as e:
-        print(
-            f"Warning: could not read config file '{config_path}': {e}",
-            file=sys.stderr,
-        )
-        config = {}
+    config = {}
+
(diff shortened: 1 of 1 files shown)


Urvity03:
Hi! I've submitted a fix for this issue in PR #114.

Feedback and review are welcome. Thanks!

Urvity03:
Thanks for the review.

I've updated the implementation to handle `UnicodeDecodeError` alongside `json.JSONDecodeError` and restored the existence check so missing config files do not generate warnings on first run.

The changes have been pushed to this PR.

Urvity03:
Thanks for the review and the additional edge cases.

I've updated the implementation to handle "json.JSONDecodeError", "UnicodeDecodeError", and "OSError", while preserving the original behavior for missing config files.

I noticed the suggestions regarding "ValueError" and "RecursionError". I'm happy to add support for those as well if that aligns with the intended scope of this issue. Please let me know your preference, and I'll update the implementation accordingly.

Thanks!

bitflicker64:
@Urvity03 Thanks for the clear explanation. Yes, please go ahead and add `ValueError` and `RecursionError` to the exception tuple. Both can be triggered by a malformed `config.json` — `ValueError` covers Python 3.11+'s integer string conversion limit during JSON parsing, and `RecursionError` covers deeply nested structures. The original `except Exception` handled both silently, so the narrowed handler should too. The final tuple would be

 ```python
 except (json.JSONDecodeError, UnicodeDecodeError, OSError, ValueError, RecursionError) as e:
 ```

This keeps the fallback-to-defaults behavior consistent for all user-triggered config file failures while still being more explicit than a bare `except Exception`.

bitflicker64:
@dosu help here

Urvity03:
Thanks for the clarification.

I've updated the exception handling to also cover "ValueError" and "RecursionError", so malformed configuration files continue to fall back to defaults while avoiding a broad "except Exception".

The changes have been pushed to this PR.

bitflicker64 (approved):
LGTM! since its alr 4/5 its safe to merge , also u can check out other issues

PR #126, opened 2026-06-23
fix: log warning on DB error in add_reminder instead of silent pass

### LINKED ISSUE #111: [Easy] reminder.py: `except Exception: pass` in `add_reminder` silently swa…
## 📍 Location

| Field | Value |
|-------|-------|
| **File** | `termstory/reminder.py` |
| **Line** | 99 |
| **Type** | `TODO` |
| **Category** | `bug` |
| **Difficulty** | Easy |

---

## 🔍 Analysis

Let me look at the relevant code in `termstory/reminder.py` around line 99.

Let me look at the database.py to understand the `get_connection` method and the DB schema, and also look at the test file for reminders to understand the expected behavior.

I now have all the information I need to answer the question comprehensively. Let me analyze the issue.

The relevant code block is lines 84-102 in `termstory/reminder.py`. The `except Exception: pass` at line 99-100 silently swallows any database errors that occur when trying to fetch the latest session. This means:

1. If the DB query fails (e.g., table doesn't exist, connection error, schema mismatch), the exception is caught and ignored.
2. The reminder is still created, but with `session_id = None` and `project_name = "Other"` (the defaults set at lines 81-82).
3. The caller has no way to know that the DB lookup failed.

Root cause: The bare `except Exception: pass` at line 99-100 swallows all exceptions from the cursor execution and fetchone, leaving `session_id` and `project_name` at their default values without any warning or logging.

Suggested fix:
1. Replace `except Exception: pass` with either:
   - Re-raise the exception (if DB errors should be fatal)
   - Log a warning (if DB errors should be non-fatal but visible)
   - At minimum, log the error so it's not completely silent

The most appropriate fix for this use case (where the DB association is "nice to have" but not critical) would be to log a warning rather than silently swallowing the error. This way the reminder is still created, but the user/developer is informed of the issue.

Let me provide a complete answer.

## 1. Context — Relevant Code Block

`termstory/reminder.py`, lines 80–102:

```python
# Get latest session if database is provided
session_id = None
project_name = "Other"

if db is not None:
    conn = db.get_connection()
    try:
        cursor = conn.cursor()
        cursor.execute("""
            SELECT s.id, p.name
            FROM sessions s
            LEFT JOIN projects p ON s.project_id = p.id
            ORDER BY s.start_time DESC
            LIMIT 1
        """)
        row = cursor.fetchone()
        if row:
            session_id = row[0]
            project_name = row[1] or "Other"
    except Exception:   # <-- line 99: the problem
        pass            # <-- line 100: silently swallowed
    finally:
        conn.close()
```

---

## 2. Root Cause

There are two distinct problems:

**a) Silent failure.** Any exception raised by `cursor.execute(...)` or `cursor.fetchone()` — a locked DB, a missing `sessions` or `projects` table, a schema mismatch, a corrupt file — is caught by the bare `except Exception: pass` and discarded entirely. No log, no warning, no re-raise.

**b) Silent fallback.** Because `session_id` and `project_name` are pre-set to their defaults (`None` / `"Other"`) before the `try` block, the reminder is written to disk with those defaults even when the DB lookup failed. The caller (`add_reminder`) returns a `Dict` that looks perfectly valid — there is no signal that the project association was lost.

---

## 3. Suggested Fix

**File:** `termstory/reminder.py`

### Step 1 — Add a `logging` import at the top of the file

```python
# termstory/reminder.py, after the existing imports
import logging

```

### Step 2 — Replace the bare `except Exception: pass` with a logged warning

```python
    except Exception as exc:
        logger.warning(
            "add_reminder: failed to fetch latest session from DB; "
            "reminder will be created without project association. Error: %s",
            exc,
        )
```

This keeps the non-fatal behaviour (the reminder is still saved) while making the failure observable in logs, rather than disappearing silently.

### Step 3 — (Optional but recommended) Add a test for the error path

In `tests/test_reminder.py`, add a test that passes a `db` whose `get_connection()` raises (or whose cursor raises on `execute`), and assert that:
- `add_reminder` still returns a valid reminder dict.
- `project_name` is `"Other"` and `session_id` is `None`.
- A warning was emitted (use `pytest`'s `caplog` fixture).

---

## 4. Effort

**S** — The core fix is a two-line change (replace `pass` with `logger.warning(...)`). Adding the import and a regression test brings it to a few hours at most.

---

## 5. Prerequisites

A newcomer needs to understand:

- **SQLite connection lifecycle in this codebase.** `db.get_connection()` returns a `SafeConnection` (a subclass of `sqlite3.Connection`) and the caller is responsible for calling `conn.close()` — already handled by the `finally` block here.
  

- **Why the fallback defaults exist.** `session_id = None` and `project_name = "Other"` are intentional defaults for the case where no `db` is passed at all (e.g., CLI invocations without an active DB). The bug is that the same defaults are also used silently on DB *errors*, which is a different situation.
  

- **Python's `logging` module.** Using `logger.warning(...)` instead of `print` is the idiomatic way to surface non-fatal errors in library code without breaking callers.

- **The reminder data model.** Each reminder JSON object carries `session_id` and `project_name` fields; if these are wrong at creation time they cannot be corrected later without re-querying the DB, so getting them right at write time matters.

---

Jpeg-create:
## Description
Replaces the bare except Exception: pass at line 99 in add_reminder with a logger.warning(...) call that captures the exception and includes context, keeping the non-fatal fallback behaviour (session_id=None, project_name="Other") but making DB failures visible in logs.
Added import logging and a module-level logger = logging.getLogger(__name__) per the maintainer's note in the issue.
Also added a regression test test_add_reminder_logs_warning_on_db_error in tests/test_reminder.py that passes a mock DB whose cursor raises, then asserts the reminder is still returned with default values and that a warning was logged via caplog.
All 6 reminder tests pass locally.

Fixes #111

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.

### DIFF
--- a/termstory/reminder.py
+++ b/termstory/reminder.py
@@ -1,9 +1,12 @@
 import json
+import logging
 import os
 import time
 import re
 from typing import List, Dict, Optional, Tuple
 from termstory.config import get_app_dir
+
+logger = logging.getLogger(__name__)
 
 def get_reminders_file_path() -> str:
     """Return path to reminders JSON file"""
@@ -96,8 +99,12 @@ def add_reminder(
             if row:
                 session_id = row[0]
                 project_name = row[1] or "Other"
-        except Exception:
-            pass
+        except Exception as exc:
+            logger.warning(
+                "add_reminder: failed to fetch latest session from DB; "
+                "reminder will be created without project association. Error: %s",
+                exc,
+            )
         finally:
             conn.close()
 
--- a/tests/test_reminder.py
+++ b/tests/test_reminder.py
@@ -77,6 +77,39 @@ def test_add_and_complete_reminder(tmp_path, monkeypatch):
     # Try completing non-existent
     assert complete_reminder(999) is False
 
+def test_add_reminder_logs_warning_on_db_error(tmp_path, monkeypatch, caplog):
+    """When the DB lookup raises, the reminder is still saved with defaults
+    and a warning is logged. Regression test for issue #111."""
+    reminders_file = tmp_path / "reminders.json"
+    monkeypatch.setattr("termstory.reminder.get_reminders_file_path", lambda: str(reminders_file))
+
+    class BrokenCursor:
+        def execute(self, *args, **kwargs):
+            raise RuntimeError("simulated DB failure")
+
+    class BrokenConn:
+        def cursor(self):
+            return BrokenCursor()
+        def close(self):
+            pass
+
+    class BrokenDB:
+        def get_connection(self):
+            return BrokenConn()
+
+    with caplog.at_level("WARNING", logger="termstory.reminder"):
+        rem = add_reminder("review code in 2 days", db=BrokenDB())
+
+    # Reminder is still created with default fallback values
+    assert rem["about"] == "review code"
+    assert rem["days"] == 2
+    assert rem["session_id"] is None
+    assert rem["project_name"] == "Other"
+
+    # Warning was emitted with the simulated error context
+    assert any("add_reminder" in r.message and "simulated DB failure" in r.message
+               for r in caplog.records)
+
 def test_cli_remind_commands(tmp_path, monkeypatch):
     reminders_file = tmp_path / "reminders.json"
     monkeypatch.setattr("termstory.reminder.get_reminders_file_path", lambda: str(reminders_file))


Jpeg-create:
CI failure on Python 3.11 is in test_tui_worker_cancellation (tests/test_tui.py), which is unrelated to this PR. Python 3.9, 3.10, and 3.12 all pass, and this PR only touches reminder.py and test_reminder.py. It looks like a flaky asyncio timing test in the TUI suite. Could you re-run the failing job to confirm?

bitflicker64:
> CI failure on Python 3.11 is in test_tui_worker_cancellation (tests/test_tui.py), which is unrelated to this PR. Python 3.9, 3.10, and 3.12 all pass, and this PR only touches reminder.py and test_reminder.py. It looks like a flaky asyncio timing test in the TUI suite. Could you re-run the failing job to confirm?

LGTM!! Thanks for your contribution

PR #128, opened 2026-06-23
fix: log warning on DB errors in ai.py context lookups instead of silent pass

### LINKED ISSUE #109: [Easy] ai.py: Bare `except Exception: return []` silently swallows al…
## 📍 Location

| Field | Value |
|-------|-------|
| **File** | `termstory/ai.py` |
| **Line** | 67 |
| **Type** | `TODO` |
| **Category** | `bug` |
| **Difficulty** | Easy |

---

## 🔍 Analysis

Let me look at the relevant code in `termstory/ai.py` around line 67.

I have all the information I need to answer this question comprehensively. Let me also check if there's a logging setup in the codebase to suggest the right logger to use.

I have enough information to answer the question. Let me compile the answer.

---

## 1. Context — Relevant Code Block

`termstory/ai.py`, lines 55–71:

```python
def _get_all_active_project_contexts() -> List[tuple]:
    conn = None
    try:
        from termstory.config import get_db_path
        from termstory.database import Database
        db_path = get_db_path()
        db = Database(db_path)
        conn = db.get_connection()
        cursor = conn.cursor()
        cursor.execute("SELECT name, project_context FROM projects WHERE project_context IS NOT NULL AND project_context != ''")
        rows = cursor.fetchall()
        return rows
    except Exception:       # ← line 67: bare catch, no logging
        return []
    finally:
        if conn is not None:
            conn.close()
```

The function is called exactly once, in `generate_timeframe_summary`, where its result is silently used as an empty list if anything goes wrong:

---

## 2. Root Cause

`except Exception: return []` catches **every possible failure** — a missing DB file, a locked DB, a schema mismatch (e.g. `project_context` column not yet migrated), a bad `get_db_path()` return value, an `ImportError`, anything — and discards both the exception object and any diagnostic information. The caller (`generate_timeframe_summary`) receives an empty list and silently omits all project contexts from the LLM prompt, with no indication that anything went wrong.

The companion function `_get_project_context_from_db` has the same pattern (`except Exception: pass`) at line 48, so this is a systemic issue in the file.

There is no `logging` import anywhere in `ai.py`, so there is currently no mechanism to surface these failures even in debug mode.

---

## 3. Suggested Fix

**Step 1 — Add a module-level logger at the top of `termstory/ai.py`** (after the existing imports):

```python
import logging
```

**Step 2 — Replace the bare `except` in `_get_all_active_project_contexts` (line 67)** with a logged, typed catch:

```python
    except Exception as e:
        logger.warning(
            "_get_all_active_project_contexts: failed to load project contexts "
            "from DB, returning empty list. Error: %s",
            e,
            exc_info=True,
        )
        return []
```

**Step 3 — Apply the same fix to `_get_project_context_from_db` (line 48)** for consistency:

```python
    except Exception as e:
        logger.warning(
            "_get_project_context_from_db: failed to load context for project %r. "
            "Error: %s",
            project_name,
            e,
            exc_info=True,
        )
```

**Step 4 (optional but recommended)** — If you want callers to be able to distinguish "DB unavailable" from "no contexts exist", consider raising a custom exception instead of returning `[]`, and let `generate_timeframe_summary` decide whether to proceed or surface an error to the UI.

---

## 4. Effort

**S** — The core fix (steps 1–3) is a mechanical, localized change: add one import, replace two bare `except` clauses with logged versions. No schema changes, no new dependencies, no test infrastructure changes required.

---

## 5. Prerequisites

- **Python `logging` module basics** — understanding `getLogger(__name__)`, log levels (`DEBUG`/`WARNING`/`ERROR`), and `exc_info=True` (which attaches the full traceback to the log record).
- **SQLite locking behavior** — the DB can raise `sqlite3.OperationalError: database is locked` under concurrent access; knowing this helps you decide whether to retry vs. log-and-return.
- **The `project_context` column migration** — this column is added via `ALTER TABLE` in `Database.init_db()` and may not exist on older installs, which is one realistic trigger for the silent failure.
  
- **How `generate_timeframe_summary` uses the result** — the contexts are injected into the LLM prompt as a `contexts_block` string; an empty list means the LLM gets no project context, producing a less personalized summary with no visible error to the user.

---

Jpeg-create:
## Description
Replaces two bare except Exception blocks in termstory/ai.py with logger.warning(..., exc_info=True) calls so DB lookup failures during project context fetching are visible in logs instead of being silently discarded.
Changes:

Added import logging and a module-level logger = logging.getLogger(__name__)
_get_project_context_from_db (line 48): replaced except Exception: pass with a logged warning. Still returns None on failure to preserve existing callers' fallback behaviour.
_get_all_active_project_contexts (line 67): replaced except Exception: return [] with a logged warning followed by return []. Still returns an empty list to preserve existing callers' fallback behaviour.

Added two regression tests in tests/test_ai_error_surfacing.py that pass mock DBs whose cursors raise, then assert the functions return their safe defaults (None and []) and that warnings were logged via caplog.
All 10 tests in test_ai_error_surfacing.py pass locally.

Fixes #109

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.

### DIFF
--- a/termstory/ai.py
+++ b/termstory/ai.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import urllib.request
 import urllib.error
 import threading
@@ -7,6 +8,7 @@ import socket
 from typing import List, Optional, Dict
 from termstory.sanitizer import sanitize_session_commands
 
+logger = logging.getLogger(__name__)
 _local_ai_state = threading.local()
 
 # Circuit Breaker Configuration
@@ -46,7 +48,12 @@ def _get_project_context_from_db(project_name: str) -> Optional[str]:
         if row:
             return row[0]
     except Exception:
-        pass
+        logger.warning(
+            "_get_project_context_from_db: failed to fetch project context for %r; "
+            "returning None.",
+            project_name,
+            exc_info=True,
+        )
     finally:
         if conn is not None:
             conn.close()
@@ -65,6 +72,11 @@ def _get_all_active_project_contexts() -> List[tuple]:
         rows = cursor.fetchall()
         return rows
     except Exception:
+        logger.warning(
+            "_get_all_active_project_contexts: failed to fetch project contexts; "
+            "returning empty list.",
+            exc_info=True,
+        )
         return []
     finally:
         if conn is not None:
--- a/tests/test_ai_error_surfacing.py
+++ b/tests/test_ai_error_surfacing.py
@@ -206,4 +206,61 @@ def test_http_error_whitespace_normalization(monkeypatch):
     assert res is None
     assert get_last_ai_error() == "HTTP Error 500: Something went wrong."
 
+def test_get_project_context_logs_warning_on_db_error(monkeypatch, caplog):
+    """When the DB lookup raises in _get_project_context_from_db,
+    the function returns None and logs a warning. Regression test for issue #109."""
+    from termstory.ai import _get_project_context_from_db
 
+    class BrokenCursor:
+        def execute(self, *args, **kwargs):
+            raise RuntimeError("simulated DB failure in fetch")
+
+    class BrokenConn:
+        def cursor(self):
+            return BrokenCursor()
+        def close(self):
+            pass
+
+    class BrokenDB:
+        def get_connection(self):
+            return BrokenConn()
+
+    monkeypatch.setattr("termstory.database.Database", lambda *args, **kwargs: BrokenDB())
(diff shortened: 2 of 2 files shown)


bitflicker64:
LGTM !!

bitflicker64 (approved):

PR #129, opened 2026-06-24
Improve zsh history file read error handling

Urvity03:
## Summary

Improve error handling in `parse_zsh_history()` by replacing the broad `except Exception` block with explicit `OSError` handling and warning logging.

## Changes

* Added module-level logging support.
* Replaced broad exception handling with `except OSError as e`.
* Added warning log output when a Zsh history file cannot be read.
* Preserved the existing fallback behavior of returning an empty command list.

## Benefits

* Avoids silently swallowing file read failures.
* Makes troubleshooting easier through warning logs.
* Keeps behavior consistent while improving observability.

### DIFF
--- a/termstory/parser.py
+++ b/termstory/parser.py
@@ -1,9 +1,12 @@
 import os
 import re
+import logging
 from datetime import datetime, timedelta
 from typing import List, Optional, Dict, Any, Union, Callable
 from termstory.models import Command
 from termstory.timestamp_detective import TimestampDetective
+
+logger = logging.getLogger(__name__)
 
 try:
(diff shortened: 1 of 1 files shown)


Urvity03:
Thanks for the review.

The current implementation follows the issue description by replacing the broad `except Exception` with explicit `OSError`handling and warning logging while preserving the existing fallback behavior.

I'm happy to make further adjustments if the maintainer would like the exception handling expanded to cover additional cases.

Dreamstick9:
LGTM, thanks for the fix!! @Urvity03

bitflicker64 (approved):
LGTM

Dreamstick9 (approved):

PR #131, opened 2026-06-24
fix: build venv in-place instead of staging+mv pattern

bitflicker64:
The staging+mv approach caused 'Too many levels of symbolic links' on macOS when the moved venv's symlinks conflicted (`bin/python3` → `bin/python3.9`). It also required fragile shebang rewriting that kept hitting edge cases (wrapper scripts, non-UTF-8 binaries, no trailing newline, `set -e` gotcha).

New approach:
1. Backup old venv to a temp dir
2. Build new venv directly at `~/.termstory-venv`
3. Install and verify
4. On any failure — restore old venv from backup
5. On success — delete backup

No `mv` of the venv itself → symlinks correct from creation → no shebang rewriting needed → zero of the previous edge cases.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -88,15 +88,20 @@ install_venv() {
     return 1
   fi
 
-  # Atomically swap — only now do we destroy the old one
-  rm -rf "$final_venv"
+  # Atomically swap: keep the old venv as a rollback target
+  local old_backup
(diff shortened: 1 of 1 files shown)


bitflicker64:
@dosu is there a way to give u power to merge?

PR #133, opened 2026-06-24
fix: migrate FTS5 virtual tables whose column set drifted from current schema

bitflicker64:
Older databases can have FTS5 virtual tables whose column set no longer matches the current code. CREATE VIRTUAL TABLE IF NOT EXISTS leaves drifted schemas intact, so triggers and INSERTs hit:

```
OperationalError: table sessions_fts has no column named ai_summary
```

This is the runtime error that was blocking ingestion on first launch.

## Fix

`init_db` introduces `_rebuild_fts_if_drifted(table, triggers, expected_cols)` which:
1. Inspects the actual columns via `pragma_table_info('table')`.
2. Drops the FTS table + its triggers if columns don't match.
3. Lets existing CREATE IF NOT EXISTS repopulate.

Guards now cover all four FTS tables in `_migrate_fts5`:
- `search_index` ⟶ {content, type, ref_id, project_id, timestamp}
- `commands_fts` ⟶ {command, exit_code}
- `sessions_fts` ⟶ {ai_summary}
- `ai_summaries_fts` ⟶ {summary}

## Why it took multiple rounds

PR #133 went through several revisions flagged by Greptile and Dosu:

**Round 1**: Subtraction bug — code did `{cols} - {'content'}` as if 'content' were a shadow column, but FTS5 declares it as a real column. A schema with both 'content' and 'ai_summary' would have slipped past the guard.

**Round 2 / Dosu**: Same drift risk on commands_fts and ai_summaries_fts; no guard for search_index.

**Round 3 / Greptile T-Rex**: commands_fts guard was placed AFTER its CREATE/INSERT block, so a drifted schema hit 'no column named exit_code' during population before the rebuild ever ran.

**Round 4 / manual test**: search_index guard for the same ordering issue caused CREATE IF NOT EXISTS to no-op, leaving the rebuilt table absent.

**Final**: All four drift checks now run before any CREATE/INSERT. Dead except branch removed (pragma_table_info returns empty set, not error).

## Verification

Forced each of four drift states in a sandbox DB copy and ran `init_db`:

| Drift state | Result |
|--------------|--------|
| sessions_fts = content+session_id+project_id+timestamp | ✓ migrated to ai_summary |
| sessions_fts = content+ai_summary (Greptile case) | ✓ migrated to ai_summary |
| commands_fts = command (missing exit_code) | ✓ migrated to command+exit_code |
| search_index = content+type (drifted) | ✓ migrated to full set |

### DIFF
--- a/termstory/database.py
+++ b/termstory/database.py
@@ -1237,6 +1237,36 @@ class Database:
         if not any('FTS5' in opt for opt in options):
             return
 
+        # FTS5 migration: older DBs may have drifted FTS column sets (e.g. sessions_fts
+        # used to have [content, session_id, project_id, timestamp]; current schema is
+        # [ai_summary]). CREATE VIRTUAL TABLE IF NOT EXISTS leaves drifted tables intact,
+        # so triggers/INSERTs hit "no such column" errors at runtime. Check actual
+        # column sets against expected and rebuild on drift. UNINDEXED is just a
+        # modifier — pragma_table_info still returns those columns.
+        #
+        # All drift checks MUST run before any FTS CREATE/INSERT below — otherwise
+        # a rebuild leaves the table absent and a subsequent CREATE IF NOT EXISTS
+        # has already run as a no-op.
+        def _rebuild_fts_if_drifted(table: str, triggers: list, expected_cols: set) -> None:
+            actual_cols = {r[0] for r in cursor.execute(
+                f"SELECT name FROM pragma_table_info('{table}')"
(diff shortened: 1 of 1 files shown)


bitflicker64:
@dosu iss the problem greptile suggested in any other places of the codebase ?

bitflicker64:
@dosu check

bitflicker64:
@greptile-apps check out dosu's comment

PR #144, opened 2026-06-24
docs: privacy documentation update — secret redaction parity fix

bitflicker64:
## What this PR does
Documentation companion to the code fix in #<insert code fix PR number here>. No source logic changes — only documentation, comments, and docstrings.

## Why each file changed

**`DATA_PRIVACY.md`** had two factual errors after the code fix lands:
1. Only mentioned "commands" being sent to LLMs — commit messages were omitted entirely from the trust story.
2. Said blacklisted sessions cause TermStory to "abort the AI request entirely" — the real behavior is replacing the COMMANDS block, not aborting. Session metadata still goes through. Fixed.
Also: named all three AI surfaces (auto-summaries, chronicle, ask), added missing blacklist token patterns, added `.termstoryignore` section.

**`docs/privacy.md`** had the same blacklist description inaccuracy, a BLACKLIST_PATTERNS code block that only showed 4 of the 8 actual patterns in `sanitizer.py`, and a redaction table with 7 of 17 implemented pattern types. All expanded to match real source.

**`SECURITY.md`** created from scratch — GitHub displays it in the Security tab and it's the standard location for vulnerability disclosure and honest known-limitations documentation. The "known gaps" table explicitly documents what the entropy heuristic misses, so users can make informed decisions rather than assuming total coverage.

**`CONTRIBUTING.md`** — the root cause of this bug was that sanitization is opt-in per function. Added a section that states the rule explicitly, shows the code pattern, and points to the existing test as a template. Future contributors now have no excuse for missing it.

**`docs/ai-integration.md`** — described the LLM transport layer but said nothing about sanitization. Added a "Pre-LLM Sanitization" section so the data flow is fully documented.

**`docs/cli-reference.md`** — `termstory ask` had no privacy disclosure in its description.

**`sanitizer.py` comments** — two inline notes about the entropy heuristic's floor (so developers don't treat it as a guarantee) and the `.termstoryignore` no-live-reload behavior (so no one files a bug when their edits don't immediately take effect).

**Docstrings** (`ask.py`, `ai.py`) — existing docstrings had zero mention of the security behavior. Updated to include the sanitization contract so it's visible at the function definition.

## Testing
No code logic changes — no new tests needed. All existing tests pass unchanged (58/58 in tested modules).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.1] - 2026-06-25
+
+### Security
+- **Secret redaction parity for `termstory ask` and Daily Chronicle**: `generate_answer()` in `ask.py` and `generate_daily_chronicle_prompt()` in `ai.py` now run session commands and git commit messages through the same local sanitizer (`sanitize_session_commands` / `redact_command`) already used by `generate_ai_summary()`. Previously, both functions built LLM prompts from completely raw, unredacted shell history and commit text.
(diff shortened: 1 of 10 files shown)


bitflicker64:
@miku review

bitflicker64:
@miku /review

PR #162, opened 2026-06-25
fix: catch specific ValueError instead of broad Exception in shlex split

### LINKED ISSUE #148: [Easy] tui.py: Bare except Exception swallows all errors during shlex …
## Location

| Field | Value |
|-------|-------|
| **File** | `termstory/tui.py` |
| **Line** | 2652 |
| **Type** | `FIXME` |
| **Category** | `bug` |
| **Difficulty** | Easy |

---

## Analysis

Let me read the relevant section of the file to analyze the FIXME comment.

The FIXME comment isn't in the file itself — the user is describing the issue conceptually. Let me read a bit more context around the area to understand the full function, and also check how shlex is imported.

I have enough context to provide a thorough analysis. Let me structure the answer.

The issue is at lines 2650-2653 in `termstory/tui.py`:

```python
try:
    shlex_tokens = shlex.split(cmd_str)
except Exception:
    shlex_tokens = tokens
```

The problem:
1. `except Exception` is too broad — it catches everything including `ValueError` (which `shlex.split` raises on malformed input like unmatched quotes), but also any other unexpected exceptions.
2. `shlex.split` only raises `ValueError` for tokenization errors (e.g., unmatched quotes). Using `except Exception` means any other unexpected error (e.g., `TypeError`, `AttributeError`) would also be silently swallowed and fall back to `tokens`, hiding bugs.
3. The correct fix is to catch only `ValueError`, which is the specific exception `shlex.split` raises for tokenization errors.

The FIXME comment (described by the user, not literally in the file) is accurate: the bare `except Exception` swallows all errors, not just the expected tokenization errors.

=== CONTEXT ===

From `termstory/tui.py`, lines 2649–2658:

```python
if base in editor_executables:
    try:
        shlex_tokens = shlex.split(cmd_str)
    except Exception:          # ← line 2652 — the problem
        shlex_tokens = tokens
    files = [t for t in shlex_tokens[1:] if not t.startswith('-')]
    for f in files:
        fname = os.path.basename(f)
        if fname:
            file_counts[fname] += 1
```

---

=== ROOT CAUSE ===

`shlex.split()` raises `ValueError` specifically when it encounters malformed shell syntax (e.g. an unmatched quote like `vim file\ 'unterminated`). The `except Exception` clause is far too broad: it silently catches **any** exception — including `TypeError`, `AttributeError`, or bugs introduced in future refactors — and falls back to the naive `cmd_str.split()` result stored in `tokens`. This means real programming errors in this block are swallowed without any log entry or re-raise, making them invisible during development and production. The fallback to `tokens` is a reasonable recovery strategy for the `ValueError` case alone, but applying it to all exceptions hides unrelated failures.

---

=== SUGGESTED FIX ===

- **FILE:** `termstory/tui.py`, **LINE: 2652** — Replace `except Exception:` with `except ValueError:` so only the documented `shlex.split` tokenization error is caught and recovered from; all other exceptions propagate normally.

```python
# Before
except Exception:
    shlex_tokens = tokens

# After
except ValueError:
    shlex_tokens = tokens
```

Optionally, add a log line inside the except block so malformed commands are observable:

```python
except ValueError:
    # cmd_str contained unmatched quotes or invalid shell syntax
    shlex_tokens = tokens
```

---

=== EFFORT ===

**S** — a single one-line change.

---

=== PREREQUISITES ===

None. The fix requires only basic Python exception-handling knowledge and awareness that `shlex.split` documents `ValueError` as its error type for bad input.

AK-Lmn:
## Description
This PR resolves issue #148 by catching a specific `ValueError` instead of using a broad `except Exception:` block during `shlex.split()`. This prevents the code from swallowing unrelated runtime errors or bugs.

Fixes #148

### DIFF
--- a/termstory/tui.py
+++ b/termstory/tui.py
@@ -2658,7 +2658,7 @@ class TermStoryWorkspace(App):
                 if base in editor_executables:
                     try:
                         shlex_tokens = shlex.split(cmd_str)
-                    except Exception:
+                    except ValueError:
                         shlex_tokens = tokens
                     files = [t for t in shlex_tokens[1:] if not t.startswith('-')]
                     for f in files:


bitflicker64:
@miku /ask help them with the greptile score

bitflicker64:
ahh timing related issue i gotta fix that

bitflicker64:
@miku /ask help them with the greptile score

bitflicker64:
@miku thank them for there contribution

bitflicker64:
@miku /merge --force

PR #164, opened 2026-06-25
Read AI provider settings from nested config

### LINKED ISSUE #154: [Medium] cli.py: Hardcoded Ollama API base URL should be configurable
## Location

| Field | Value |
|-------|-------|
| **File** | `termstory/cli.py` |
| **Line** | 703 |
| **Type** | `FIXME` |
| **Category** | `design` |
| **Difficulty** | Medium |

---

## Analysis

Let me read the relevant file to find the FIXME comment at line 703.

The FIXME comment isn't literally in the file, but the user is pointing to line 703-704 where `api_base_url = "http://localhost:11434/v1"` is hardcoded. Let me look at more context — the config system, any existing env var handling, and the broader function to give a complete analysis.

Let me gather more context in parallel - the config.py file, the tui.py ollama references, and the docs/configuration.md to understand the full picture.

Now I have enough context to provide a thorough analysis. Let me look at the tui.py ollama references and the cli.py environ usages to complete the picture, but I think I have enough to answer already.

Key findings:
1. Line 703-704 in `cli.py`: `elif provider == "ollama": api_base_url = "http://localhost:11434/v1"` — this is the hardcoded fallback
2. BUT in `config.py` lines 155-159, the `load_config()` function already defines a default for `providers.ollama.api_base_url` as `"http://localhost:11434/v1"` in the defaults dict, and `merge_defaults` will fill it in if missing
3. The config lookup at line 695 in cli.py: `api_base_url = config.get(f"providers.{provider}.api_base_url", config.get(f"{provider}_api_base_url", ""))` — this uses a custom `config.get()` method

Wait, `config` is a plain dict returned by `load_config()`. So `config.get(f"providers.{provider}.api_base_url", ...)` would be calling `dict.get()` with a dot-separated key like `"providers.ollama.api_base_url"` — but that's NOT how dict.get works. It would look for a literal key `"providers.ollama.api_base_url"` in the top-level dict, which doesn't exist. The nested structure uses actual nested dicts.

So the real bug is: `config.get("providers.ollama.api_base_url", ...)` always returns `""` (the default) because `dict.get()` doesn't traverse dot-separated paths. The `get_config_value()` function in `config.py` does handle dot-separated paths, but it's not being used here.

This means:
1. `api_base_url` will always be `""` at line 695 (since `dict.get` doesn't handle dot paths)
2. The `if not api_base_url:` block at line 698 always triggers
3. The hardcoded fallback at line 704 is what actually sets the Ollama URL

The fix should be:
1. Use `get_config_value(config, f"providers.{provider}.api_base_url")` instead of `config.get(f"providers.{provider}.api_base_url", ...)` at line 695
2. Remove the hardcoded fallback block at lines 698-704 (since `load_config()` already merges defaults including the Ollama URL)

OR alternatively, since `load_config()` already merges defaults (including `providers.ollama.api_base_url`), if the dot-path lookup were fixed, the fallback block would be redundant anyway.

Let me also check if there's an `OLLAMA_HOST` env var convention that should be supported.

Actually, let me re-read line 695 more carefully:
```python
api_base_url = config.get(f"providers.{provider}.api_base_url", config.get(f"{provider}_api_base_url", ""))
```

`config` is a plain `dict`. `dict.get("providers.ollama.api_base_url", ...)` looks for the literal string key `"providers.ollama.api_base_url"` in the top-level dict — which doesn't exist. So it falls back to `config.get("ollama_api_base_url", "")` — also doesn't exist. So `api_base_url = ""`.

Then `if not api_base_url:` is True, and the hardcoded fallback kicks in.

The root cause is that `config.get()` (plain dict method) is used instead of `get_config_value(config, ...)` which handles dot-separated paths. Since `load_config()` already stores the default `"http://localhost:11434/v1"` in `config["providers"]["ollama"]["api_base_url"]`, the hardcoded fallback at line 704 is a workaround for the broken lookup — but it's also not configurable via env var.

The FIXME is about the hardcoded URL, but the deeper issue is the broken config lookup.

Let me also check tui.py to see if it has the same pattern or does it correctly.

I have enough information to provide a thorough analysis. Let me synthesize everything.

Key findings:
1. Line 703-704 in `cli.py`: `elif provider == "ollama": api_base_url = "http://localhost:11434/v1"` — hardcoded fallback
2. `config.py` `load_config()` already defines `providers.ollama.api_base_url = "http://localhost:11434/v1"` as a default in the nested dict structure
3. Line 695 in `cli.py`: `api_base_url = config.get(f"providers.{provider}.api_base_url", config.get(f"{provider}_api_base_url", ""))` — uses plain `dict.get()` which does NOT traverse dot-separated paths
4. `config.py` has `get_config_value(config, path)` which DOES handle dot-separated paths, but it's not used in cli.py for this lookup
5. Because `dict.get("providers.ollama.api_base_url")` always returns `""`, the `if not api_base_url:` block always fires, and the hardcoded fallback is what actually sets the URL

The FIXME is pointing at the symptom (hardcoded URL), but the root cause is the broken config lookup using `dict.get()` with dot-separated keys instead of `get_config_value()`.

The fix:
1. `termstory/cli.py`, line 694-695: Replace `config.get(f"providers.{provider}.api_key", ...)` and `config.get(f"providers.{provider}.api_base_url", ...)` with `get_config_value(config, f"providers.{provider}.api_key")` etc. — need to import `get_config_value` from `termstory.config`
2. `termstory/cli.py`, lines 698-704: Remove the hardcoded fallback block (since `load_config()` already merges defaults including the Ollama URL, and with a correct lookup it will be found)

Optionally, also support `OLLAMA_HOST` env var as a third-tier override.

Effort: S (a few hours)

=== CONTEXT ===

`termstory/cli.py`, lines 692–713:

```python
if ai_enabled and provider != "disabled":
    from termstory.ai import translate_git_anger
    api_key = config.get(f"providers.{provider}.api_key", config.get(f"{provider}_api_key", ""))
    api_base_url = config.get(f"providers.{provider}.api_base_url", config.get(f"{provider}_api_base_url", ""))
    model_name = config.get(f"providers.{provider}.model_name", config.get(f"{provider}_model_name", ""))

    if not api_base_url:
        if provider == "groq":
            api_base_url = "https://api.groq.com/openai/v1"
        elif provider == "openai":
            api_base_url = "https://api.openai.com/v1"
        elif provider == "ollama":
            api_base_url = "http://localhost:11434/v1"   # ← line 704

    translation = translate_git_anger(
        commit_data,
        api_key=api_key,
        api_base_url=api_base_url,
        ...
    )
```

---

=== ROOT CAUSE ===

There are actually **two layered problems**, not one.

**Problem 1 (the deeper bug):** `config` is a plain `dict` returned by `load_config()`. Calling `config.get("providers.ollama.api_base_url", ...)` uses Python's built-in `dict.get()`, which looks for the *literal string key* `"providers.ollama.api_base_url"` at the top level of the dict — it does not traverse nested dicts. That key never exists, so the call always returns `""`. `config.py` already provides `get_config_value(config, path)` which correctly traverses dot-separated paths through nested dicts, but it is never imported or called in `cli.py` for these lookups.

**Problem 2 (the FIXME symptom):** Because the lookup always yields `""`, the `if not api_base_url:` block always fires, and the hardcoded string `"http://localhost:11434/v1"` is the *only* way the Ollama URL ever gets set. A user who runs Ollama on a different host or port (e.g. a remote server, a Docker container, or a non-default port) cannot override it — not through the config file, not through an environment variable.

Note: `load_config()` already stores `"http://localhost:11434/v1"` as the default for `providers.ollama.api_base_url` in the nested config dict, so the hardcoded fallback block is entirely redundant once the lookup is fixed.

---

=== SUGGESTED FIX ===

- **FILE: `termstory/cli.py`, LINE: 693** — Add `get_config_value` to the import from `termstory.config` (currently only `load_config` is imported on line 687).

- **FILE: `termstory/cli.py`, LINES: 694–696** — Replace the three `config.get(f"providers.{provider}.*", ...)` calls with `get_config_value(config, f"providers.{provider}.api_key")`, `get_config_value(config, f"providers.{provider}.api_base_url")`, and `get_config_value(config, f"providers.{provider}.model_name")` respectively. Fall back to `""` if the result is `None`.

- **FILE: `termstory/cli.py`, LINES: 698–704** — Remove the entire `if not api_base_url:` fallback block. The defaults are already baked into `load_config()` and will be returned correctly once the lookup is fixed.

- **FILE: `termstory/cli.py`, LINE: 698** *(optional but recommended)* — If env-var override is desired (e.g. for Docker/CI), add `api_base_url = os.environ.get("OLLAMA_HOST", api_base_url)` after the lookup, consistent with how `config.py` already uses `os.environ` for other settings.

---

=== EFFORT ===

**S** — The core fix is a one-line import change and replacing three `dict.get()` calls with `get_config_value()`. The optional env-var support adds one more line. No schema changes, no migrations needed.

---

=== PREREQUISITES ===

- Understand that Python's `dict.get("a.b.c")` does **not** traverse nested dicts — it looks for a literal key. This is the central misconception the bug rests on.
- Familiarity with how `load_config()` merges defaults in `termstory/config.py` (lines 140–166 and 236–250), so you know the defaults are already present and the fallback block is safe to delete.
- Know that `get_config_value()` in `termstory/config.py` is the correct utility for dot-path lookups and is already used elsewhere in the codebase (e.g. `tui.py`).

andrewkernel:
## Summary
- read CLI AI provider settings with `get_config_value` so nested provider config is honored
- remove duplicated hardcoded API base URL fallbacks from the AI CLI paths
- preserve legacy flat-key fallbacks for older configs
- add regression coverage for nested Ollama config and legacy fallback behavior

Closes #154.

## Verification
- `.\.venv\Scripts\python.exe -m pytest tests\test_git_blame_anger_fortune_teller.py -q`
- `.\.venv\Scripts\python.exe -m pytest tests\test_git_blame_anger_fortune_teller.py -q -k ai_provider`
- `.\.venv\Scripts\python.exe -m pytest tests\test_git_blame_anger_fortune_teller.py tests\test_ask.py -q`
- `python -m py_compile termstory\cli.py tests\test_git_blame_anger_fortune_teller.py`
- `git diff --check`

Note: the full `tests` suite currently stops during collection on Windows due to unrelated temp-file locking in `tests/test_bash_inspect.py` and `tests/test_null_bytes_zsh.py`.

### DIFF
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -7,7 +7,7 @@ from dateutil import parser as date_parser
 # Supported tags for session categorization
 TAGS = ["deploy", "debug", "setup", "test", "docs"]
 
-from termstory.config import get_history_files, get_db_path
+from termstory.config import get_config_value, get_history_files, get_db_path
 from termstory.parser import parse_all_histories
 from termstory.session import create_sessions
 from termstory.project import detect_projects
@@ -54,6 +54,23 @@ def safe_init_db(db: Database) -> None:
                 f"Could not initialize database: {e}"
             )
             sys.exit(1)
+
+def get_ai_provider_settings(config: dict) -> tuple[str, str, str, str]:
+    """Return active AI provider settings from nested config with legacy fallback."""
+    provider = config.get("active_provider") or config.get("ai_provider", "disabled")
+    api_key = (
+        get_config_value(config, f"providers.{provider}.api_key")
+        or config.get(f"{provider}_api_key", "")
+    )
+    api_base_url = (
+        get_config_value(config, f"providers.{provider}.api_base_url")
+        or config.get(f"{provider}_api_base_url", "")
+    )
+    model_name = (
+        get_config_value(config, f"providers.{provider}.model_name")
+        or config.get(f"{provider}_model_name", "")
+    )
+    return provider, api_key, api_base_url, model_name
 
 def intercept_sys_argv():
     """Intercept positional date arguments (e.g. termstory 2026-06-02) and rewrite them
@@ -687,21 +704,10 @@ def anger_translator(
     from termstory.config import load_config
     config = load_config()
     ai_enabled = config.get("ai_enabled", False)
-    provider = config.get("ai_provider", "disabled")
+    provider, api_key, api_base_url, model_name = get_ai_provider_settings(config)
     
     if ai_enabled and provider != "disabled":
         from termstory.ai import translate_git_anger
-        api_key = config.get(f"providers.{provider}.api_key", config.get(f"{provider}_api_key", ""))
-        api_base_url = config.get(f"providers.{provider}.api_base_url", config.get(f"{provider}_api_base_url", ""))
-        model_name = config.get(f"providers.{provider}.model_name", config.get(f"{provider}_model_name", ""))
-        
-        if not api_base_url:
-            if provider == "groq":
-                api_base_url = "https://api.groq.com/openai/v1"
-            elif provider == "openai":
-                api_base_url = "https://api.openai.com/v1"
-            elif provider == "ollama":
-                api_base_url = "http://localhost:11434/v1"
                 
         console.print("[bold yellow]Translating developer's git blame and terminal frustration...[/]")
         translation = translate_git_anger(
@@ -746,21 +752,10 @@ def fortune_teller(
     from termstory.config import load_config
     config = load_config()
     ai_enabled = config.get("ai_enabled", False)
-    provider = config.get("ai_provider", "disabled")
+    provider, api_key, api_base_url, model_name = get_ai_provider_settings(config)
     
     if ai_enabled and provider != "disabled":
         from termstory.ai import predict_bugs_from_sessions
-        api_key = config.get(f"providers.{provider}.api_key", config.get(f"{provider}_api_key", ""))
-        api_base_url = config.get(f"providers.{provider}.api_base_url", config.get(f"{provider}_api_base_url", ""))
-        model_name = config.get(f"providers.{provider}.model_name", config.get(f"{provider}_model_name", ""))
-        
-        if not api_base_url:
-            if provider == "groq":
-                api_base_url = "https://api.groq.com/openai/v1"
-            elif provider == "openai":
-                api_base_url = "https://api.openai.com/v1"
-            elif provider == "ollama":
-                api_base_url = "http://localhost:11434/v1"
                 
         console.print("[bold yellow]Foretelling potential bugs from late-night chaotic session telemetry...[/]")
         predictions = predict_bugs_from_sessions(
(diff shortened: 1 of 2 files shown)


bitflicker64:
wait the ci have some problem i am going ahead and merging !!

PR #166, opened 2026-06-25
Narrow timestamp date parse exceptions

### LINKED ISSUE #145: [Easy] timestamp_detective.py: Bare except Exception swallows all errors during git lo…
## Location

| Field | Value |
|-------|-------|
| **Line** | 1024 |
| **Type** | `FIXME` |
| **Category** | `bug` |
| **Difficulty** | Easy |

---

## Analysis

Let me read the specific file and look at the context around line 1024.

The FIXME comment isn't in the file itself — it was described by the user as a known issue. Let me look at the actual code around line 1024 more carefully. I have the context already. Let me also look at what exceptions `datetime.strptime` can raise, and check if there are any related patterns elsewhere in the file for comparison.

The code at line 1024 uses `except Exception:` which is a broad catch. The issue is:
1. `datetime.strptime` raises `ValueError` for format mismatches — that's the expected exception to catch when trying multiple formats.
2. Using `except Exception:` also silently swallows unexpected errors like `TypeError`, `AttributeError`, `OverflowError`, etc., making debugging very hard.
3. The `continue` means any unexpected error is silently ignored and the function returns `None`, giving no indication of what went wrong.

Let me also check the broader file for context on error handling patterns used elsewhere.

I have enough context to provide a thorough analysis. Let me now write the structured answer.

The issue at line 1024 is `except Exception:` which is too broad. When iterating over date format strings and trying `datetime.strptime`, the only expected failure is `ValueError` (format mismatch). Using `except Exception:` also silently swallows:
- `TypeError` (if `date_str` is not a string)
- `OverflowError` (if the timestamp is out of range)
- `OSError` (from `.timestamp()` on some platforms for out-of-range dates)
- Any other unexpected runtime error

The `continue` means all these errors are silently ignored, and the function returns `None` with no diagnostic information. This makes debugging very hard — if there's a bug in the parsing logic (e.g., wrong slice index on line 1020), it would be silently swallowed.

The fix is straightforward: replace `except Exception:` with `except ValueError:` on line 1024. Same issue exists at line 1047 in `_parse_last_login`.

Let me provide the structured analysis.

=== CONTEXT ===

```python
# termstory/timestamp_detective.py, lines 1011–1026
def _parse_git_log_date(self, log_output: str) -> Optional[int]:
    """
    Parse the first `Date:` line out of `git log` (or `brew log`) output and
    return it as a Unix timestamp.  Handles both the default git date format
    and the ISO (`--date=iso`) format.
    """
    for line in log_output.splitlines():
        stripped = line.strip()
        if stripped.startswith("Date:"):
            date_str = stripped[len("Date:"):].strip()
            for fmt in ("%a %b %d %H:%M:%S %Y %z", "%Y-%m-%d %H:%M:%S %z"):
                try:
                    return int(datetime.strptime(date_str, fmt).timestamp())
                except Exception:   # <-- line 1024: the problem
                    continue
    return None
```

---

=== ROOT CAUSE ===

`datetime.strptime` raises `ValueError` when the input string does not match the given format — that is the only exception that should be caught here, because it signals "try the next format." Using `except Exception:` instead silently swallows every other runtime error that could arise on that same line: a `TypeError` if `date_str` is somehow not a string (e.g., the slice on line 1020 misbehaves), an `OverflowError` or `OSError` from `.timestamp()` if the parsed date is out of the platform's representable range, or any future regression introduced nearby. All of those failures are masked by the bare `continue`, causing the function to return `None` with no diagnostic signal — making bugs in this code path effectively invisible.

The same anti-pattern is repeated in `_parse_last_login` at line 1047.

---

=== SUGGESTED FIX ===

- **FILE:** `termstory/timestamp_detective.py`, **LINE: 1024** — change `except Exception:` → `except ValueError:` so only a format-mismatch is silently skipped; all other exceptions propagate normally.
- **FILE:** `termstory/timestamp_detective.py`, **LINE: 1047** — same change: `except Exception:` → `except ValueError:` in `_parse_last_login` for the same reason.

If `.timestamp()` out-of-range errors (`OSError`/`OverflowError`) are a realistic concern for the date ranges this code handles, line 1024 can be widened slightly to `except (ValueError, OverflowError, OSError):` — but that should be an explicit, documented decision, not a blanket `Exception`.

---

=== EFFORT ===

**S** — both changes are single-token substitutions. The only real work is deciding whether `OverflowError`/`OSError` from `.timestamp()` need to be included and adding a brief comment explaining the choice.

---

=== PREREQUISITES ===

- Familiarity with Python's `datetime.strptime` and the exceptions it raises (`ValueError` on format mismatch).
- Understanding that `.timestamp()` can raise `OSError` or `OverflowError` on some platforms for dates outside the Unix epoch representable range (relevant if the git log could contain very old or far-future dates).
- No special environment setup needed; the change is pure Python with no external dependencies.

### LINKED ISSUE #146: [Easy] timestamp_detective.py: Bare except Exception swallows all errors during last l…
## Location

| Field | Value |
|-------|-------|
| **Line** | 1047 |
| **Type** | `FIXME` |
| **Category** | `bug` |
| **Difficulty** | Easy |

---

## Analysis

Let me read the specific file and look at the context around line 1047.

The FIXME comment isn't in the file itself — it was provided by the user as a description of the issue. Let me read a bit more context around the function to understand the full picture, including the function signature and class context.

I have enough context to provide a thorough analysis. The FIXME comment was provided by the user as a description of the issue, not actually in the file. Let me analyze the code around line 1047.

The issue is clear: `except Exception:` on line 1047 catches all exceptions (including things like `MemoryError`, `SystemExit`, `KeyboardInterrupt` — well, those are `BaseException`, not `Exception`, but the point is it catches too broadly). More specifically, the bare `except Exception` swallows the error silently — there's no logging, no re-raise, no differentiation between a `ValueError` (which is what `datetime.strptime` raises on a bad format string) and other unexpected exceptions.

The correct fix is to narrow the exception to `ValueError`, which is what `datetime.strptime` raises when the format doesn't match. This way, unexpected errors (like `AttributeError`, `TypeError`, etc.) would propagate up rather than being silently swallowed.

Let me provide the structured analysis.

=== CONTEXT ===

From `termstory/timestamp_detective.py`, lines 1038–1053:

```python
m = re.search(
    r'([A-Z][a-z]{2}\s+[A-Z][a-z]{2}\s+\d{1,2}\s+\d{1,2}:\d{2})',
    last_output
)
if not m:
    return None
date_str = re.sub(r'\s+', ' ', m.group(1)).strip()
try:
    dt = datetime.strptime(date_str, "%a %b %d %H:%M")
except Exception:          # <-- line 1047: the problem
    return None
now = datetime.now()
dt = dt.replace(year=now.year)
```

=== ROOT CAUSE ===

`datetime.strptime` raises `ValueError` when the input string does not match the format — that is the only exception it is documented to raise under normal conditions. Using `except Exception` instead of `except ValueError` means any other unexpected exception (e.g., a `TypeError` caused by a bug introduced later, an `AttributeError` from a refactor, or even an `OverflowError` from the `.replace(year=...)` call on line 1050 if it were moved inside the block) would be silently caught and discarded as `None`. This makes bugs invisible: the caller gets back `None` and has no way to distinguish "the `last` output had no parseable date" from "something went wrong in the parsing logic itself." There is also no logging at any level, so failures leave no trace.

=== SUGGESTED FIX ===

- **FILE:** `termstory/timestamp_detective.py`, **LINE: 1047** — Replace `except Exception:` with `except ValueError:` so only the expected strptime format-mismatch error is caught. All other exceptions will propagate normally.

- **FILE:** `termstory/timestamp_detective.py`, **LINE: 1047** — Optionally add a `logging.debug(...)` or `logging.warning(...)` call inside the except block (before `return None`) to record the bad `date_str` value, aiding future diagnosis without breaking callers.

The same pattern appears at lines 1022–1025 in the function just above `_parse_last_login`, where `except Exception: continue` is used inside a format-trial loop — that one is more defensible (trying multiple formats intentionally), but narrowing it to `except ValueError: continue` there is also good hygiene.

=== EFFORT ===

**S** — The core change is a one-word substitution (`Exception` → `ValueError`). Adding a log line is another one-liner. The same pattern fix at line 1024 is equally trivial.

=== PREREQUISITES ===

- Familiarity with Python's `datetime.strptime` and the exceptions it raises (`ValueError` on format mismatch).
- Understanding that `except Exception` catches everything derived from `Exception` (but not `BaseException` subclasses like `SystemExit`/`KeyboardInterrupt`), and why that is too broad when only one specific failure mode is expected.
- None of the macOS-specific `last` command knowledge is needed to make this change safely.

andrewkernel:
## Summary
- catch only `ValueError` for git log date format fallbacks
- catch only `ValueError` for macOS `last` timestamp parsing
- add regression coverage that preserves format-mismatch handling while surfacing unexpected parser errors

Closes #145.
Closes #146.

## Verification
- `.\.venv\Scripts\python.exe -m pytest tests\test_timestamp_detective.py::TestDetectMacosSystemLog::test_parse_git_log_date_iso tests\test_timestamp_detective.py::TestDetectMacosSystemLog::test_parse_git_log_date_ignores_format_mismatches_only tests\test_timestamp_detective.py::TestDetectMacosSystemLog::test_parse_last_login_ignores_format_mismatches_only -q`
- `python -m py_compile termstory\timestamp_detective.py tests\test_timestamp_detective.py`
- `git diff --check`

Note: the full `tests\test_timestamp_detective.py` file still has unrelated Windows path expectation failures in virtual-CWD tests in this checkout.

### DIFF
--- a/termstory/timestamp_detective.py
+++ b/termstory/timestamp_detective.py
@@ -1021,7 +1021,7 @@ class TimestampDetective:
                 for fmt in ("%a %b %d %H:%M:%S %Y %z", "%Y-%m-%d %H:%M:%S %z"):
                     try:
                         return int(datetime.strptime(date_str, fmt).timestamp())
-                    except Exception:
+                    except ValueError:
                         continue
         return None
 
@@ -1044,7 +1044,7 @@ class TimestampDetective:
         date_str = re.sub(r'\s+', ' ', m.group(1)).strip()
         try:
             dt = datetime.strptime(date_str, "%a %b %d %H:%M")
-        except Exception:
+        except ValueError:
             return None
         now = datetime.now()
         dt = dt.replace(year=now.year)
--- a/tests/test_timestamp_detective.py
+++ b/tests/test_timestamp_detective.py
@@ -34,6 +34,12 @@ from termstory.timestamp_detective import TimestampDetective
 # Helpers
 # ---------------------------------------------------------------------------
 
+class ParserBugDateTime:
+    @staticmethod
+    def strptime(*_args, **_kwargs):
+        raise TypeError("unexpected parser bug")
+
+
 NOW = int(datetime.now().timestamp())
 FOUR_YEARS_AGO = NOW - (4 * 365 * 24 * 60 * 60)
 SIX_YEARS_AGO  = NOW - (6 * 365 * 24 * 60 * 60)
@@ -857,6 +863,20 @@ class TestDetectMacosSystemLog(unittest.TestCase):
         iso = dt.astimezone().strftime("%Y-%m-%d %H:%M:%S %z")
         ts = self.d._parse_git_log_date(f"commit x\nDate:   {iso}\n\n    msg\n")
         self.assertEqual(ts, FOUR_YEARS_AGO)
+
+    def test_parse_git_log_date_ignores_format_mismatches_only(self):
+        self.assertIsNone(self.d._parse_git_log_date("commit x\nDate:   not a date\n"))
+
+        with patch("termstory.timestamp_detective.datetime", ParserBugDateTime):
+            with self.assertRaises(TypeError):
+                self.d._parse_git_log_date("commit x\nDate:   not a date\n")
+
+    def test_parse_last_login_ignores_format_mismatches_only(self):
+        self.assertIsNone(self.d._parse_last_login("alice ttys000 Tue Jun 99 09:14 still logged in"))
+
+        with patch("termstory.timestamp_detective.datetime", ParserBugDateTime):
+            with self.assertRaises(TypeError):
+                self.d._parse_last_login("alice ttys000 Thu Jun 5 09:14 still logged in")
 
     def test_detect_macos_syslog_missing_file(self):
         """When install.log doesn't exist (e.g. on Linux), return None."""


bitflicker64 (approved):
LGTMM!!!!

bitflicker64:
/ approve /lgtm /merge --force

bitflicker64:
@miku /chat ask them to star the repo

bitflicker64:
@miku /chat ask them to star the repo

PR #168, opened 2026-06-26
Replace broad exception handling when reading install log

### LINKED ISSUE #147: [Easy] timestamp_detective.py: Bare except Exception swallows all errors when reading …
## Location

| Field | Value |
|-------|-------|
| **Line** | 1072 |
| **Type** | `FIXME` |
| **Category** | `bug` |
| **Difficulty** | Easy |

---

## Analysis

Let me read the specific file and look at the context around line 1072.

The FIXME comment isn't in the file itself - it was provided by the user as a description of the issue. Let me read a bit more context around line 1072 to understand the full picture, and also check if there are any related patterns elsewhere.

I now have enough context to provide a thorough analysis. Let me structure the answer.

The issue is at line 1072:
```python
except Exception:
    return None
```

This is a bare `except Exception:` (no variable binding, no logging) that silently swallows all errors when opening/reading the macOS install log file. The problem is:

1. It catches ALL exceptions (including `PermissionError`, `OSError`, `UnicodeDecodeError`, etc.) without logging or re-raising them.
2. The caller has no way to distinguish between "file doesn't exist" (handled at line 1067-1068) and "file exists but couldn't be read due to permissions or other errors."
3. Silent failure makes debugging very hard.

The fix would be to catch only specific, expected exceptions (like `OSError`, `PermissionError`) and log a warning for unexpected ones, rather than silently swallowing everything.

Note: there's also a second bare `except Exception:` at line 1102, but that one is inside the per-line parsing loop and uses `continue` rather than `return None`, which is more acceptable (it skips unparseable lines). The FIXME specifically calls out line 1072.

Let me provide the structured analysis.

=== CONTEXT ===

From `termstory/timestamp_detective.py`, lines 1066–1083:

```python
log_path = "/private/var/log/install.log"
if not os.path.exists(log_path):
    return None
try:
    with open(log_path, "r", encoding="utf-8", errors="ignore") as f:
        lines = f.readlines()
except Exception:          # ← line 1072  FIXME
    return None

needle = package.lower()
date_re = re.compile(r'^(\d{4}-\d{2}-\d{2}[ T]\d{2}:\d{2}:\d{2})...')
best_ts: Optional[int] = None
for line in lines:
    low = line.lower()
    if "installed" not in low:
        continue
```

---

=== ROOT CAUSE ===

The `except Exception:` block on line 1072 catches every possible exception — `PermissionError`, `OSError`, `MemoryError`, `KeyboardInterrupt` (well, not that one, but anything derived from `Exception`) — and silently discards it with a bare `return None`. This means:

1. **Silent failure**: a `PermissionError` (very common for `/private/var/log/install.log` on macOS without elevated privileges) is indistinguishable from a clean "file not found" path. The caller in `_run_all_detectors` gets `None` and moves on with no indication anything went wrong.
2. **No logging**: there is no `logger.warning(...)` or `logger.debug(...)` call, so the error is completely invisible during normal operation and even during debugging.
3. **Overly broad catch**: only `OSError` (which covers `PermissionError`, `FileNotFoundError`, `IOError`) is actually expected here. Catching all of `Exception` also hides programming errors like `AttributeError` or `TypeError` that should surface as bugs.

There is a second `except Exception: continue` at line 1102 inside the per-line date-parsing loop; that one is more defensible (skipping a malformed line), but it also lacks any logging. 

---

=== SUGGESTED FIX ===

- **FILE:** `termstory/timestamp_detective.py`, **LINE: 1072** — Replace `except Exception:` with `except OSError as exc:` and add a `logger.debug("Could not read %s: %s", log_path, exc)` (or `logger.warning`) before `return None`, so permission/IO failures are visible in logs without crashing the pipeline.

- **FILE:** `termstory/timestamp_detective.py`, **LINE: 1102** — Similarly replace `except Exception:` with `except (ValueError, OverflowError) as exc:` (the only real exceptions `datetime.fromisoformat` / `datetime.strptime` / `int()` can raise here) and optionally add a `logger.debug(...)` call before `continue`.

---

=== EFFORT ===

**S** — Both changes are one-liners plus an optional log call. The only care needed is verifying which exception types `datetime.fromisoformat` and `open()` actually raise in the relevant Python versions.

---

=== PREREQUISITES ===

- Familiarity with Python exception hierarchy (`OSError` vs `Exception`).
- Know that `/private/var/log/install.log` requires `sudo` or the `admin` group on macOS Ventura+, so `PermissionError` is the most common real-world failure here.
- If the project uses a structured logger (check for a `logger = logging.getLogger(...)` at the top of the file), use that; otherwise `import logging` and add one.

Urvity03:
## Summary

This PR replaces the overly broad `except Exception` block in `termstory/timestamp_detective.py` with `except OSError` when reading the macOS install log.

## Changes

* Replaced `except Exception:` with `except OSError:`
* Preserved the existing behavior by returning `None` when the install log cannot be read
* Kept the change focused and limited to the reported issue

Fixes #147

### DIFF
--- a/termstory/timestamp_detective.py
+++ b/termstory/timestamp_detective.py
@@ -1069,7 +1069,7 @@ class TimestampDetective:
         try:
             with open(log_path, "r", encoding="utf-8", errors="ignore") as f:
                 lines = f.readlines()
-        except Exception:
+        except OSError:
             return None
 
         needle = package.lower()


bitflicker64:
@miku /review

bitflicker64:
@miku /approve

bitflicker64:
@miku /lgtm

bitflicker64:
@miku /merge

bitflicker64:
@miku /shrug

bitflicker64:
@miku /merge --forcce

bitflicker64 (approved):

bitflicker64:
@miku /merge --force

bitflicker64:
@miku /chat thank them for contributing

PR #180, opened 2026-06-29
test(tui): add docstring to is_worker_cancelled

bitflicker64:
Test PR for the new Miku auto area-label feature.

Touches only `termstory/tui.py` so `.miku.yml` should auto-apply **just** `area/tui` and nothing else. Clean isolated test.

If this works, the bot should add the `area/tui` label within a few seconds of opening — no `:miku /classify` needed.


bitflicker64:
:miku /help

PR #203, opened 2026-07-02
fix(date_utils): chain exception in parse_date_range_helper

### LINKED ISSUE #197: [Easy] date_utils.py: `except Exception: raise ValueError(...)` in `parse_dat…
## Summary
The `parse_date_range_helper` function in `termstory/date_utils.py` uses a bare `except Exception:` clause at line 108 that discards the original parsing exception when raising a `ValueError`, preventing proper exception chaining and making debugging harder.

## Steps to Reproduce
1. Call `parse_date_range_helper` with an invalid date string that triggers a `dateutil.parser.parse` error (e.g., a malformed date format)
2. Observe that the raised `ValueError` does not include the original exception in its traceback

## Expected vs Actual
**Expected**: The `ValueError` raised on line 109 should chain the original `dateutil` parsing exception using `from e`, similar to the pattern used at lines 99-100, so developers can see the root cause in the traceback.

**Actual**: The bare `except Exception:` on line 108 does not bind the exception to a variable, and the `raise ValueError` on line 109 does not use `from e`, so the original parsing error is silently discarded from the exception chain.

## Environment
Not specified

## Context
This issue affects the `parse_date_range_helper` function which is used by the `web` command to parse date range filters like `--date-range today` or `--date-range 2026-06-01:2026-06-15`. The function is also tested in `test_parse_date_range_helper`. The fix is a two-line change to match the exception chaining pattern already used elsewhere in the same function.

Maqbool61:
## Summary
This PR improves exception handling in `parse_date_range_helper` by adding exception chaining (`from e`) to preserve the original `dateutil` parsing error in the traceback when date parsing fails. The change applies to both the colon-range parsing branch and the single-date fallback path, making error handling consistent across the function and improving debuggability by maintaining the full error context. 

## Changes
- **Core**: Updated exception handling in `parse_date_range_helper` in `termstory/date_utils.py` to use `raise ValueError(...) from e` at both error-raise sites:
  - Colon-range branch (line 100): Added `from e` to preserve the original `dateutil` parse error when colon-separated date parsing fails 
  - Single-date fallback path (lines 108-109): Changed `except Exception:` to `except Exception as e:` and added `from e` to preserve the original parse error when single-date parsing fails 
- **Tests**: Added `test_parse_date_range_helper_chains_underlying_error` in `tests/test_date_utils.py` to verify that the `ValueError` raised by `parse_date_range_helper` correctly chains the underlying `dateutil` parse error via `__cause__` 

## Fixes
- Fixes #197 — Improves exception handling in date parsing to preserve original error context

## Breaking Changes
None

## Testing
Run the date utils tests to verify the changes:
```bash
pytest tests/test_date_utils.py -v
```
The existing `test_parse_date_range_flexible` test covers the `parse_date_range_helper` function with various date formats, and the new `test_parse_date_range_helper_chains_underlying_error` test specifically verifies the exception chaining behavior. 

## Notes
- This PR standardizes exception handling across both parsing branches in `parse_date_range_helper`, ensuring that the original `dateutil` parsing errors are always preserved in the traceback for easier debugging.
- The review thread initially identified an inconsistency where the colon-range branch lacked `from e` chaining; this PR resolves that by adding exception chaining to both branches.

### DIFF
--- a/termstory/date_utils.py
+++ b/termstory/date_utils.py
@@ -97,7 +97,7 @@ def parse_date_range_helper(date_range_str: str) -> Tuple[int, int]:
             end = datetime.combine(end_dt.date(), time.max)
             return int(start.timestamp()), int(end.timestamp())
         except Exception as e:
-            raise ValueError(f"Invalid date range format: {e}")
+            raise ValueError(f"Invalid date range format: {e}") from e
             
     # Try parsing as a single date
     try:
@@ -105,6 +105,6 @@ def parse_date_range_helper(date_range_str: str) -> Tuple[int, int]:
         start = datetime.combine(dt.date(), time.min)
         end = datetime.combine(dt.date(), time.max)
         return int(start.timestamp()), int(end.timestamp())
-    except Exception:
-        raise ValueError(f"Unknown date range format '{date_range_str}'")
+    except Exception as e:
+        raise ValueError(f"Unknown date range format '{date_range_str}'") from e
 
--- a/tests/test_date_utils.py
+++ b/tests/test_date_utils.py
@@ -1,3 +1,4 @@
+import pytest
 import os
 from datetime import datetime
 from termstory.date_utils import get_current_time, get_today_range, get_week_range, get_month_range, format_date_range
@@ -96,3 +97,10 @@ def test_timezone_aware_override(monkeypatch):
     monkeypatch.setenv("TERMSTORY_DATE_OVERRIDE", "2026-06-19T12:00:00+05:30")
     ct = get_current_time()
     assert ct.tzinfo is None
+
+def test_parse_date_range_helper_chains_underlying_error():
+    """Invalid input must raise ValueError with the original parse error chained via __cause__."""
+    from termstory.date_utils import parse_date_range_helper
+    with pytest.raises(ValueError) as exc:
+        parse_date_range_helper("not-a-real-date-at-all-zzzzz")
+    assert exc.value.__cause__ is not None, "ValueError must chain the underlying parse error"


termstory/date_utils.py
```diff
@@ -105,6 +105,6 @@ def parse_date_range_helper(date_range_str: str) -> Tuple[int, int]:
         start = datetime.combine(dt.date(), time.min)
         end = datetime.combine(dt.date(), time.max)
         return int(start.timestamp()), int(end.timestamp())
-    except Exception:
-        raise ValueError(f"Unknown date range format '{date_range_str}'")
+    except Exception as e:
```
bitflicker64, line 108: **Inconsistency to address.** Greptile P2 + my review both flagged this — the colon-range branch at **line 100** (directly above this one) should also chain the cause for the two error paths to behave consistently.

```python
        except Exception as e:
            raise ValueError(f"Invalid date range format: {e}") from e
```

Without `from e` here, that branch silently swallows the underlying `dateutil` error while the one this PR fixes preserves it — so callers debugging a colon-range parse failure see only "Unknown date range format" instead of the actual dateutil message.

**Out of scope for an inline suggestion** since line 100 isn't in this PR's diff hunks. Two options:

1. Add the same `from e` here in this PR (smallest, keeps it consistent in one commit)
2. Land this as-is and file a follow-up issue to address line 100

I'd lean toward option 1 — it's a 1-line change and the PR description was originally written to claim it matches this branch, so fixing the claim is cleaner than leaving a known inconsistency in the tree.

---

**Optional test addition** for `tests/test_date_utils.py` to lock in the chain behavior:

```python
import pytest


def test_parse_date_range_helper_chains_underlying_error():
    """Invalid input must raise ValueError with the original parse error chained via __cause__."""
    from termstory.date_utils import parse_date_range_helper
    with pytest.raises(ValueError) as exc:
        parse_date_range_helper("not-a-real-date-at-all-zzzzz")
    assert exc.value.__cause__ is not None, "ValueError must chain the underlying parse error"
```

No test currently asserts `__cause__` is preserved — a future refactor could silently drop `from e` and nothing would notice.

Thanks for the fix! :miku /merge once the colon-range branch is consistent.

Maqbool61: working on fix

bitflicker64 (approved):
_LGTM_

Dreamstick9 (approved):

PR #206, opened 2026-07-02
[AI] Share LLM error state across threads

### LINKED ISSUE #122: [Medium] ai.py: thread-local state hides background thread errors
## Summary
The thread-local storage used for AI error state in `termstory/ai.py` prevents error messages from background worker threads from being visible to the main caller thread, causing `get_last_ai_error()` to always return `None` after background failures. 

## Steps to Reproduce
1. Trigger an AI summary generation in the TUI (e.g., by selecting a session and generating a story) 
2. Cause the background LLM request to fail (e.g., network error, invalid API key, timeout)
3. The background `_worker` thread sets the error in thread-local storage 
4. The main thread calls `get_last_ai_error()` to retrieve the error for display 
5. Observe that `get_last_ai_error()` returns `None` instead of the actual error message

## Expected vs Actual
**Expected**: The caller thread should be able to retrieve the most recent error message from the background worker thread via `get_last_ai_error()`, allowing the TUI to display meaningful error notifications to the user. 

**Actual**: Since `_local_ai_state` uses `threading.local()`, each thread has its own copy of the error state. The background worker thread sets the error in its thread-local storage, but when the main caller thread invokes `get_last_ai_error()`, it reads from its own separate thread-local copy, which is empty, always returning `None`. 

## Environment
Not specified

## Context
The issue was introduced in version 0.2.11 when thread-safe AI error tracking was added using thread-local storage to support concurrent background workers.  However, the implementation fails to account for the fact that the TUI's background workers (decorated with `@work(thread=True)`) run in separate threads from the main caller thread that needs to read the errors. 

The suggested fix is to replace the thread-local storage with a shared `queue.Queue(maxsize=1)` or a single value protected by `threading.Lock` to ensure the caller always sees the most recent worker failure. A test should also be added to verify that errors scheduled in the worker are correctly read by the caller. 

## Notes
The existing test `test_concurrent_threads_error_isolation` in `tests/test_ai_error_surfacing.py` actually tests the opposite behavior—it verifies that errors are isolated between threads, which is the current (buggy) behavior that needs to be changed.

sjsjsjjs534:
## Summary
- Replace thread-local AI error storage with a lock-protected shared last-error value
- Ensure failures from LLM requests made on background request threads are visible to callers
- Update concurrent error-surfacing coverage for the shared latest-error behavior

## Tests
- .venv/bin/python -m pytest tests/test_ai.py -k last_ai_error_is_shared_across_request_threads -q
- XDG_CONFIG_HOME=/private/tmp/termstory-test-config XDG_DATA_HOME=/private/tmp/termstory-test-data .venv/bin/python -m pytest tests/test_ai.py tests/test_ai_error_surfacing.py -q
- XDG_CONFIG_HOME=/private/tmp/termstory-test-config XDG_DATA_HOME=/private/tmp/termstory-test-data .venv/bin/python -m pytest tests/ -q

Closes #122

### DIFF
--- a/termstory/ai.py
+++ b/termstory/ai.py
@@ -9,7 +9,10 @@ from typing import List, Optional, Dict
 from termstory.sanitizer import sanitize_session_commands, redact_command
 
 logger = logging.getLogger(__name__)
+# Per-thread error state — each caller (archive, dashboard, etc.)
+# keeps its own last error so switching screens doesn't lose context.
 _last_ai_error_lock = threading.Lock()
+_local_ai_state = threading.local()
 _last_ai_error: Optional[str] = None
 
 # Circuit Breaker Configuration
--- a/termstory/ai.py
+++ b/termstory/ai.py
@@ -9,7 +9,8 @@ from typing import List, Optional, Dict
 from termstory.sanitizer import sanitize_session_commands, redact_command
 
 logger = logging.getLogger(__name__)
-_local_ai_state = threading.local()
+_last_ai_error_lock = threading.Lock()
+_last_ai_error: Optional[str] = None
 
 # Circuit Breaker Configuration
 _circuit_breaker_lock = threading.Lock()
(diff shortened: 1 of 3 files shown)


bitflicker64:
:miku /chat tell them to star the repo

termstory/ai.py
```diff
@@ -9,7 +9,8 @@
 from termstory.sanitizer import sanitize_session_commands, redact_command
 
 logger = logging.getLogger(__name__)
-_local_ai_state = threading.local()
+_last_ai_error_lock = threading.Lock()
```
bitflicker64, line 14:
```suggestion
# Per-thread error state — each caller (archive, dashboard, etc.)
# keeps its own last error so switching screens doesn't lose context.
_last_ai_error_lock = threading.Lock()
_local_ai_state = threading.local()
```

Replacing `threading.local()` with a single global `_last_ai_error` means all feature screens now share one error state. If the archive screen fails ("HTTP 503") and then the daily summary screen succeeds, `clear_last_ai_error()` at the start of the second call clears the archive screen's error. The user sees no error on the archive screen even though their AI request failed.

The old thread-local approach was correct per-caller isolation. The new global introduces a cross-contamination bug where one screen's success masks another screen's failure. Suggest keeping the thread-local for error isolation while using `_last_ai_error_lock` for the circuit-breaker state (which IS genuinely shared).

bitflicker64:
@copilot resolve the merge conflicts in this pull request

PR #210, opened 2026-07-03
fix: make NFS timeout cache TTL configurable via nfs_timeout_cache_ttl (closes #199)

### LINKED ISSUE #199: [Medium] project.py: Hardcoded 60-second TTL for the timed-out NFS path cach…
## Summary

The `_listdir_with_timeout` function in `termstory/project.py` uses a hardcoded 60-second TTL for caching timed-out NFS paths, preventing users from configuring this timeout value without modifying source code.

## Steps to Reproduce

Not applicable - this is a design/feature request.

## Expected vs Actual

**Expected**: The NFS timeout cache TTL should be configurable via `config.json` (e.g., `nfs_timeout_cache_ttl`), following the same pattern as other tunable values like `db_timeout` and `ai_cooldown_seconds`.

**Actual**: The TTL is hardcoded as the literal `60` on line 107 of `termstory/project.py`, with no mechanism for users to adjust it without editing the source code.

## Environment

Not specified.

## Context

The hardcoded value appears in the `_listdir_with_timeout` function which implements a safety mechanism for network mounts. When a directory listing times out (e.g., on a hung NFS mount), the path is cached in `_timed_out_paths` and subsequent calls are short-circuited for 60 seconds to prevent repeated hangs.

The codebase already has an established pattern for configurable timeouts: values like `db_timeout` (30.0), `ai_cooldown_seconds` (60.0), and `reminder_poll_interval` (300) are defined in the `defaults` dict within `load_config()` and can be modified via `termstory config set <key> <value>`.

The fix should follow this same pattern by:
1. Adding a module-level default constant in `termstory/project.py`
2. Reading the value from `load_config()` with a fallback to the default
3. Adding the key to the `defaults` dict in `termstory/config.py`
4. Adding test coverage to verify the config value is respected (extending `test_listdir_timeout_caching` in `tests/test_project.py`)

## Notes

The existing test `test_listdir_timeout_caching` validates that the cache prevents repeated hangs by checking that a second timeout call completes in under 0.05 seconds. This test should be extended to verify that a custom TTL from config is respected instead of the hardcoded 60 seconds.

Maqbool61:
## Problem

In `termstory/project.py`, the `_listdir_with_timeout` function caches paths that have timed out on NFS mounts to avoid repeated hangs. However, the TTL (time-to-live) for this cache was hardcoded as `60` seconds:

if now - _timed_out_paths[path] < 60:

This meant users had no way to adjust the cache duration without modifying the source code directly. Fixes #199.

---

### 1. `termstory/config.py`
Added `nfs_timeout_cache_ttl` to the `defaults` dict inside `load_config()` so it becomes a first-class config option with a sensible default of `60` seconds:

"nfs_timeout_cache_ttl": 60,

### 2. `termstory/project.py`
Replaced the hardcoded `60` in `_listdir_with_timeout` with a dynamic read from `load_config()`. If the config cannot be loaded for any reason, it safely falls back to `60`:

from termstory.config import load_config
try:
    ttl = load_config().get("nfs_timeout_cache_ttl", 60)
except Exception:
    ttl = 60

if now - _timed_out_paths[path] < ttl:

### 3. `tests/test_project.py`
Added a new test `test_listdir_timeout_caching_custom_ttl` that verifies the custom TTL from config is actually respected:

- First call times out and the path gets cached
- Second call immediately raises `TimeoutError` from cache (confirms caching works)
- After waiting for the custom TTL (1 second) to expire, a third call bypasses the cache and attempts `listdir` again (confirms TTL expiry works)

---

## Test Results

All 19 tests pass with no regressions:

19 passed in 2.49s

---

## How to Configure

Users can now set a custom TTL in their config.json:

{
    "nfs_timeout_cache_ttl": 120
}

This controls how long (in seconds) a timed-out NFS path stays cached before being retried. Default remains `60` seconds for full backward compatibility.

### DIFF
--- a/termstory/config.py
+++ b/termstory/config.py
@@ -173,7 +173,8 @@ def load_config() -> dict:
         ],
         "reminder_poll_interval": 300,
         "clustering_threshold": 0.6,
-        "default_branch_names": ["main"]
+        "default_branch_names": ["main"],
+        "nfs_timeout_cache_ttl": 60,
     }
     
     config = {}
--- a/termstory/project.py
+++ b/termstory/project.py
@@ -101,10 +101,16 @@ _timeout_lock = threading.Lock()
 
 def _listdir_with_timeout(path: str, timeout: float = 0.5) -> List[str]:
     """Execute os.listdir in a daemon thread to enforce a timeout (e.g. on hung NFS mounts)"""
+    from termstory.config import load_config
+    try:
+        ttl = load_config().get("nfs_timeout_cache_ttl", 60)
+    except Exception:
+        ttl = 60
+
     now = time.time()
     with _timeout_lock:
         if path in _timed_out_paths:
-            if now - _timed_out_paths[path] < 60:
+            if now - _timed_out_paths[path] < ttl:
                 raise TimeoutError(f"os.listdir recently timed out on path (cached): {path}")
             else:
                 del _timed_out_paths[path]
--- a/termstory/project.py
+++ b/termstory/project.py
@@ -99,18 +99,21 @@ import threading
 _timed_out_paths = {}  # path -> timestamp of last timeout
 _timeout_lock = threading.Lock()
 
-def _listdir_with_timeout(path: str, timeout: float = 0.5) -> List[str]:
-    """Execute os.listdir in a daemon thread to enforce a timeout (e.g. on hung NFS mounts)"""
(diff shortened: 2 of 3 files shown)


bitflicker64:
:miku what do u think? looks good to me ?

PR #211, opened 2026-07-03
fix(ai): replace silent exception handlers with specific logging

### LINKED ISSUE #190: [Easy] ai.py: 5 bare `except Exception` blocks silently swallow errors
## Summary
`termstory/ai.py` contains 5 bare `except Exception` blocks that silently catch and suppress all errors without logging, making AI provider failures invisible to users and difficult to debug.

## Steps to Reproduce
1. Trigger a database error during project context fetch (e.g., corrupted database file)
2. Trigger a network error during LLM API call (e.g., connection refused, timeout)
3. Trigger a malformed JSON response from an LLM provider
4. Observe that no error traceback is logged and the user receives no error message

## Expected vs Actual
**Expected:** Specific exceptions should be caught (e.g., `sqlite3.DatabaseError`, `urllib.error.URLError`, `json.JSONDecodeError`, `KeyError`) and errors should be logged via `logger.exception()` so tracebacks are available at debug level for debugging.

**Actual:** All exceptions are silently caught by bare `except Exception` blocks, errors are not logged, and users see empty responses or silent fallbacks to local mode with no indication of what went wrong.

## Environment
Not specified

## Context
This is the same anti-pattern as issues #149 and #113, but occurring in the AI provider hot path within `termstory/ai.py`. The bare except blocks are located in:
- Database context fetch functions (`_get_project_context_from_db`, `_get_all_active_project_contexts`)
- The retry/backoff loop's worker thread (`_worker` inside `_send_llm_request`)
- HTTP error response parsing logic

When LLM calls fail due to rate limits, JSON parse errors, or network issues, the current implementation provides no visibility into the failure, making it impossible for users to debug configuration or connectivity problems.

Urvity03:
## Summary

This PR replaces broad `except Exception` handlers in `termstory/ai.py` with specific exception types and adds descriptive `logger.exception(...)` calls to improve error observability while preserving existing fallback behavior and control flow.

### Exception Handling Improvements
- **`termstory/ai.py`**: Added `import sqlite3` to support database-specific exception types
- **`_get_project_context_from_db`**: Replaced `except Exception:` with `except (sqlite3.DatabaseError, OSError, RuntimeError)` and added explicit `return None` after logging
- **`_get_all_active_project_contexts`**: Replaced `except Exception:` with `except (sqlite3.DatabaseError, OSError, ImportError, RuntimeError)`
- **`_worker` (in `_send_llm_request`)**: Replaced `except Exception as e:` with specific exception types `(urllib.error.URLError, OSError, UnicodeDecodeError, json.JSONDecodeError, KeyError, IndexError, TypeError, AttributeError, RuntimeError, ConnectionError)` and added `logger.exception("LLM request failed.")`
- **`_worker` HTTP error parsing**: Replaced `except Exception:` with `except (json.JSONDecodeError, ValueError, AttributeError)` and added `logger.exception("Failed to parse LLM HTTP error response JSON.")`
- **`_worker` HTTP error body reading**: Replaced `except Exception:` with `except (OSError, UnicodeDecodeError, ValueError)` and added `logger.exception("Failed to read HTTP error body from LLM response.")`
- **`generate_daily_chronicle_prompt`**: Replaced `except Exception:` with `except (OSError, UnicodeDecodeError)` and added `logger.exception("Failed to load daily chronicle template example.")`

## Fixes

- Fixes #190 — Replaces broad exception handlers with specific exception types to improve error observability

## Breaking Changes

None

## Testing

- `python -m py_compile termstory/ai.py` — Verify syntax correctness
- Manual verification that no `except Exception:` or bare `except:` handlers remain in `termstory/ai.py`
- Existing test coverage in `tests/test_ai_error_surfacing.py` validates DB error logging behavior

## Notes

- Greptile review noted that the `_worker` thread handler omits `http.client.HTTPException` subclasses that `response.read()` can raise in practice; this may be addressed in a follow-up
- All existing fallback behavior, retries, and control flow are preserved
- The changes improve observability by logging exceptions with stack traces via `logger.exception()` instead of silently swallowing errors

### DIFF
--- a/termstory/ai.py
+++ b/termstory/ai.py
@@ -3,6 +3,7 @@ import logging
 import sqlite3
 import urllib.request
 import urllib.error
+import http.client
 import threading
 import time
 import socket
@@ -137,7 +138,7 @@ def _get_all_active_project_contexts() -> List[tuple]:
         cursor.execute("SELECT name, project_context FROM projects WHERE project_context IS NOT NULL AND project_context != ''")
         rows = cursor.fetchall()
         return rows
-    except (sqlite3.DatabaseError, OSError, ImportError, RuntimeError):
+    except (sqlite3.DatabaseError, OSError, RuntimeError):
         logger.warning("_get_all_active_project_contexts: failed to fetch project contexts; returning empty list.", exc_info=True)
         return []
     finally:
@@ -241,8 +242,7 @@ def _send_llm_request(
                     if result.startswith("'") and result.endswith("'"):
                         result = result[1:-1]
                     result_box.append(result)
-            except (urllib.error.URLError, OSError, UnicodeDecodeError, json.JSONDecodeError, KeyError, IndexError, TypeError, AttributeError, RuntimeError, ConnectionError) as e:
-                logger.exception("LLM request failed.")
+            except (urllib.error.URLError, OSError, UnicodeDecodeError, json.JSONDecodeError, KeyError, IndexError, TypeError, AttributeError, RuntimeError, ConnectionError, http.client.HTTPException) as e:
                 error_box.append(e)
 
         worker_thread = threading.Thread(target=_worker, daemon=True)
@@ -343,6 +343,8 @@ def _send_llm_request(
             return escape(result_box[0])
         return None
 
+    if error_box:
+        logger.warning("LLM request failed after %d attempts: %s", max_retries + 1, error_box[-1])
     return None
 
 def generate_ai_summary(
--- a/termstory/ai.py
+++ b/termstory/ai.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import sqlite3
 import urllib.request
 import urllib.error
 import threading
@@ -9,6 +10,7 @@ from typing import List, Optional, Dict
 from termstory.sanitizer import sanitize_session_commands, redact_command
 
 logger = logging.getLogger(__name__)
+
 _local_ai_state = threading.local()
 
 # Circuit Breaker Configuration
@@ -115,13 +117,9 @@ def _get_project_context_from_db(project_name: str) -> Optional[str]:
         row = cursor.fetchone()
         if row:
             return row[0]
-    except Exception:
-        logger.warning(
-            "_get_project_context_from_db: failed to fetch project context for %r; "
-            "returning None.",
-            project_name,
-            exc_info=True,
-        )
+    except (sqlite3.DatabaseError, OSError, RuntimeError):
+        logger.warning("_get_project_context_from_db: failed to fetch project context for %r; returning None.", project_name, exc_info=True)
+        return None
     finally:
         if conn is not None:
             conn.close()
@@ -139,12 +137,8 @@ def _get_all_active_project_contexts() -> List[tuple]:
         cursor.execute("SELECT name, project_context FROM projects WHERE project_context IS NOT NULL AND project_context != ''")
         rows = cursor.fetchall()
         return rows
-    except Exception:
-        logger.warning(
-            "_get_all_active_project_contexts: failed to fetch project contexts; "
-            "returning empty list.",
-            exc_info=True,
-        )
+    except (sqlite3.DatabaseError, OSError, ImportError, RuntimeError):
+        logger.warning("_get_all_active_project_contexts: failed to fetch project contexts; returning empty list.", exc_info=True)
         return []
     finally:
         if conn is not None:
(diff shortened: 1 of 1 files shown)


bitflicker64:
@Urvity03 so what u gonna have to do is go to your main branch and add upstream ( this repo url) and then do git pull!!
after that u gonna go to your Urvity03:fix/ai-bare-exception-logging and then do a git pull ?
got it ? this is one way to do it!! and also star the repo @Dreamstick9 aint gonna merge it if u didnt lol

bitflicker64:
if u have any questions join the discord and ask -  https://discord.gg/PHcXrachS ... Works?

bitflicker64:
:miku /chat explain what i explained but better since branch has conficts

bitflicker64:
:miku /check-star

Urvity03:
:miku /check-star

bitflicker64:
@Urvity03 4/5 score from greptile You are almost there !!! 🥲😓

Dreamstick9 (approved):

bitflicker64 (changes requested):
.

bitflicker64 (changes requested):
.

bitflicker64 (changes requested):
Mostly good — but a couple things to tighten up before merge:

**L21 (`except (sqlite3.DatabaseError, OSError, ImportError)`)** — `ImportError` doesn't belong in a DB query catch block. If the module can't be imported at runtime that's a startup failure, not a query failure. This masks real bugs. Drop it or add a comment explaining why it's intentional.

**L26 (`except Exception as e` → 8-exception tuple)** — The massive tuple is correct in spirit but misses `RuntimeError` (threading errors in `_worker`) and `ConnectionError` (network layer beneath the LLM call). If those fire, they still fall through to `except Exception` at the caller. Consider adding those two or leaving a note that the remaining `except Exception` at the outer caller handles the rest.

**L33 (`logger.exception(\"Failed to parse...\")`)** — The `logger.exception()` call here is good but it's inside a nested `except` that catches the same `json.JSONDecodeError` that the outer `except Exception` would catch. This means the inner `except` fires first and the outer `except` never sees it — but the outer `except Exception` has no `logger.exception`. So if the outer fires, the user gets no log. Add `logger.exception()` to the outer too.

Otherwise solid — the pattern is correct (broad → narrow), just those three edge cases.

bitflicker64 (changes requested):
Three things to fix:

**L31** — `except (sqlite3.DatabaseError, OSError, ImportError)` — `ImportError` doesn't belong in a DB query catch. If someone removes `termstory/ai.py` at runtime, the `ImportError` gets caught silently and the user never knows. This is a module-level failure, not a query failure. Drop it or add a comment explaining why it's intentional.

**L57** — `except (urllib.error.URLError, OSError, UnicodeDecodeError, ...)` — this 8-tuple is correct in spirit but missing `RuntimeError` (threading errors in `_worker`) and `ConnectionError` (network below the LLM call). If those fire, they fall through to the `except Exception` at the outer caller — but that outer is L24-31 which has no `logger.exception`. So the user gets a silent failure. Add both to the tuple.

**L66** — `logger.exception("Failed to parse...")` is inside a nested `except` that catches `json.JSONDecodeError`. But the outer `except Exception` (L73) also catches it — the inner fires first, the outer never sees it. The outer has no `logger.exception`. So if `json.JSONDecodeError` fires here, the inner catches it and logs, but if it fires *on the outer's scope* (L62-73), the user gets no log. Add `logger.exception` to L73 or restructure the nesting.

Otherwise good — the pattern is correct (`except Exception` → narrow types), just these three edge cases.

bitflicker64:
okay my bot is not working i am infuriated

bitflicker64 (approved):
looks good

bitflicker64 (changes requested):
Three things to fix:

```suggestion
    except (sqlite3.DatabaseError, OSError):
```
L31 — `ImportError` doesn't belong in a DB query catch. If the module file is removed at runtime, `ImportError` fires silently and the user never knows. This is a startup failure, not a query failure. Drop it.

```suggestion
        except (urllib.error.URLError, OSError, UnicodeDecodeError, json.JSONDecodeError, KeyError, IndexError, TypeError, AttributeError, RuntimeError, ConnectionError) as e:
```
L57 — missing `RuntimeError` (threading errors in `_worker`) and `ConnectionError` (network below the LLM call). If those fire, they fall through to the outer bare `except Exception` which has no `logger.exception`. The user gets a silent failure. Add both.

```suggestion
                        except Exception:
                            logger.exception("Failed to parse LLM HTTP error response JSON.")
```
L66 — the inner `except json.JSONDecodeError` catches it first, but the outer `except Exception` (L73) also catches it. The inner fires first, the outer never sees it. The outer has no `logger.exception`. Add `logger.exception` to the outer or restructure the nesting so the outer also logs.

Otherwise good — the pattern is correct (broad → narrow), just these three edge cases.

bitflicker64:
sorry for the bad pr experience will be improving as a maintainer lol u can take up anyother pr 😬

PR #234, opened 2026-07-06
issue #198 : fixed query=None calls

### LINKED ISSUE #198: [Medium] rag.py: `advanced_search(db, query=None, ...)` fetches ALL sess…
## Summary
The `hybrid_search` function in `termstory/rag.py` passes `query=None` to `advanced_search`, causing it to fetch all sessions from the database instead of using FTS5 pre-filtering, resulting in O(N) database queries and expensive embedding inference on the entire dataset.

## Expected vs Actual

**Expected**: `hybrid_search` should use FTS5 pre-filtering to retrieve a bounded candidate set (e.g., top-K results) and then re-rank those candidates using BM25 and semantic embedding scores. This is the standard "retrieve-then-rerank" pattern that leverages the existing FTS5 infrastructure in `termstory/search.py`.

**Actual**: `hybrid_search` passes `query=None` to `advanced_search`, which bypasses all text-based filtering. This causes `_search_standard` to execute a `WHERE 1=1` query that returns every session in the database filtered only by metadata predicates (project, date range, tags). Then `_populate_results` fires two additional SQL queries per session (fetching all commands and commits), creating an N+1 query problem. Finally, the system performs expensive transformer model inference on every session to compute semantic embeddings, making the entire operation O(N) in session count.

## Environment
Not specified

## Context

The issue occurs in the semantic/RAG search feature introduced in version 0.6.0. When users run `termstory search --semantic`, the CLI calls `hybrid_search` which is intended to combine BM25 keyword matching with semantic embedding similarity for better relevance ranking.

The root cause is that `hybrid_search` intentionally passes `query=None` to avoid "double-filtering" by text, but this defeats the purpose of the FTS5 index. The FTS5 infrastructure (`commands_fts`, `sessions_fts`, `ai_summaries_fts`, and `search_index`) already exists and can efficiently return a small, relevance-ranked candidate set, but `hybrid_search` never uses it.

The proposed fix is to:
1. Pass the actual `query` string and `fts=True` to `advanced_search` in `rag.py`
2. Add a `limit` parameter to `_search_standard`, `_search_fts5`, and `_search_new_fts5` to bound the candidate set size
3. Add a `top_k` parameter to `hybrid_search`'s signature to control the candidate set size

This would enable the standard retrieve-then-rerank pattern where FTS5 provides fast candidate retrieval and the hybrid scorer provides high-quality re-ranking.

advik604:
## Summary
This PR optimizes the semantic/RAG search path by using FTS-backed candidate retrieval instead of pulling the full dataset, significantly improving performance for large databases. It adds a `top_k` parameter to bound the candidate set and implements a fallback mechanism that retrieves recent sessions when FTS finds no matches, ensuring semantic reranking still works. 

### Core
- **termstory/rag.py**: Added `top_k: Optional[int] = 20` parameter to `hybrid_search()` function to bound the candidate retrieval set 
- **termstory/rag.py**: Changed first `advanced_search()` call to use `query=query` with `fts=True` and `limit=top_k` for FTS-backed candidate retrieval 
- **termstory/rag.py**: Added fallback logic: if no candidates found and query is truthy, re-query with `query=None` and `limit=top_k` to get recent sessions for semantic reranking 
- **termstory/rag.py**: Updated docstring to document the FTS-backed candidate retrieval with fallback behavior 

- **termstory/search.py**: Added `limit: Optional[int] = None` parameter to `advanced_search()` function 
- **termstory/search.py**: Passed `limit` parameter to all three internal search functions: `_search_new_fts5()`, `_search_fts5()`, and `_search_standard()` 
- **termstory/search.py**: Added `limit` parameter to `_search_new_fts5()` and implemented SQL `LIMIT` clause when limit is not None 
- **termstory/search.py**: Added `limit` parameter to `_search_fts5()` and implemented SQL `LIMIT` clause when limit is not None 
- **termstory/search.py**: Added `limit` parameter to `_search_standard()` and implemented SQL `LIMIT` clause when limit is not None 

### Tests
- **tests/test_rag.py**: Updated `test_hybrid_search_success_with_mocked_dependency` to expect 1 result instead of 2 due to `top_k=20` limiting results 
- **tests/test_rag.py**: Removed hybrid_score comparison assertions since only 1 result is returned 
- **tests/test_rag.py**: Added new test `test_hybrid_search_falls_back_when_fts_has_no_matches` to verify the fallback behavior when FTS returns no matches 

## Fixes
- Fixes #198 — Semantic/RAG search performance optimization by using FTS-backed candidate retrieval instead of pulling full dataset

## Breaking Changes
None

## Testing
Run the RAG tests to verify the changes:
```bash
pytest tests/test_rag.py -v
```

Specifically, the new test validates the fallback behavior:
```bash
pytest tests/test_rag.py::test_hybrid_search_falls_back_when_fts_has_no_matches -v
```

## Notes
The fallback mechanism ensures that semantic search still works even when FTS finds no text matches for the query, by retrieving the most recent `top_k` sessions as a bounded candidate pool for cosine similarity reranking. This maintains semantic relevance while avoiding full dataset scans.

### DIFF
--- a/termstory/rag.py
+++ b/termstory/rag.py
@@ -113,7 +113,8 @@ def hybrid_search(
     until_ts: Optional[int] = None,
     tag_filters: Optional[List[str]] = None,
     alpha: float = 0.5,
-    model_name: str = "all-MiniLM-L6-v2"
+    model_name: str = "all-MiniLM-L6-v2",
+    top_k: Optional[int] = 20
 ) -> List[Dict]:
     """
     Performs a hybrid search (BM25 + Cosine Similarity) over terminal sessions.
@@ -125,15 +126,17 @@ def hybrid_search(
             "Please install it using: pip install sentence-transformers"
         )
         
-    # Fetch all candidate sessions matching metadata filters (but without the query text filter)
+    # Retrieve a bounded candidate set using the text-aware search pipeline before reranking.
     from termstory.search import advanced_search
     candidate_sessions = advanced_search(
         db,
-        query=None,
+        query=query,
         project_filter=project_filter,
         since_ts=since_ts,
         until_ts=until_ts,
-        tag_filters=tag_filters
+        tag_filters=tag_filters,
+        fts=True,
+        limit=top_k
     )
     
     if not candidate_sessions or not query:
--- a/termstory/rag.py
+++ b/termstory/rag.py
@@ -118,6 +118,8 @@ def hybrid_search(
 ) -> List[Dict]:
     """
     Performs a hybrid search (BM25 + Cosine Similarity) over terminal sessions.
+    Uses FTS-backed candidate retrieval first, but falls back to a bounded non-FTS
+    candidate set when no FTS matches are found so semantic reranking still works.
     If sentence-transformers is not installed, raises an ImportError.
     """
     if not SENTENCE_TRANSFORMERS_AVAILABLE:
@@ -138,6 +140,17 @@ def hybrid_search(
         fts=True,
         limit=top_k
     )
+
+    if not candidate_sessions and query:
+        candidate_sessions = advanced_search(
+            db,
+            query=None,
+            project_filter=project_filter,
+            since_ts=since_ts,
+            until_ts=until_ts,
+            tag_filters=tag_filters,
+            limit=top_k
+        )
     
     if not candidate_sessions or not query:
         return candidate_sessions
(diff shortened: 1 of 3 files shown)


bitflicker64:
:miku /ask does it look mergable ?

bitflicker64:
Well lgtm thanks for the contribution
 /approve /lgtm /triage /merge

PR #268, opened 2026-07-12
feat: support natural language dates for --date option

### LINKED ISSUE #255: [Medium] cli.py: `--date` accepts only strict YYYY-MM-DD; `--date-range` accepts natural-language dates — inconsistent
## Summary
The global `--date` flag in `termstory/cli.py` only accepts strict YYYY-MM-DD format, while the `--date-range` flag accepts natural-language dates like "yesterday" and "today", creating inconsistent CLI behavior.

## Steps to Reproduce
1. Run `PYTHONPATH=. python3 -m termstory --date yesterday today`
2. Observe error: `Error: Invalid date format 'yesterday'`
3. Run `PYTHONPATH=. python3 -m termstory web --date-range yesterday`
4. Observe the command succeeds (natural-language supported)

## Expected vs Actual
**Expected**: The `--date` global flag should accept the same natural-language inputs as `--date-range`, including "today", "yesterday", "tomorrow", and "Ndays" formats, plus any dateutil-parseable string.

**Actual**: The `--date` flag only accepts strict YYYY-MM-DD format because it uses raw `date_parser.parse()` instead of the `parse_date_range_helper()` function that handles natural-language parsing.

## Environment
Not specified

## Context
The `termstory/date_utils.py` module already implements `parse_date_range_helper()` which understands natural-language date formats including "today", "yesterday", "Ndays", and "YYYY-MM-DD:YYYY-MM-DD" ranges. The `web` command's `--date-range` flag successfully uses this helper, but the global `--date` flag in the CLI callback does not, causing the inconsistency.

Ayush-0918:
## What this does

Makes the global `--date` option accept natural-language inputs like `today`, `yesterday`, and `7days` — the same strings `--date-range` already supports — and normalizes the result so `TERMSTORY_DATE_OVERRIDE` always gets a clean `YYYY-MM-DD` value.

Before, `--date` only validated the string with `dateutil` and passed it through unchanged. That worked for literal dates, but failed for the shorthand forms. Now it reuses the shared `parse_date_range_helper()` and writes back a normalized date, so downstream code that reads `TERMSTORY_DATE_OVERRIDE` sees a consistent format.

## Changes

- `termstory/cli.py`
  - Added `datetime` import
  - Replaced direct `date_parser.parse()` validation with `parse_date_range_helper()`
  - Normalizes the parsed result to `YYYY-MM-DD` before setting `TERMSTORY_DATE_OVERRIDE`
  - Updated `--date` help text
- `tests/test_cli_commands.py`
  - Added `test_global_date_accepts_natural_language()` for `yesterday`
  - Added manual env cleanup so the override doesn't leak into later tests

## Fix

Closes #255.

## How to test

```bash
pytest tests/test_cli_commands.py::test_global_date_accepts_natural_language
```

Or manually:

```bash
termstory --date yesterday
termstory --date today
termstory --date 7days
termstory --date not-a-date   # should error out
```

## Notes

No breaking changes. The timezone behavior follows the existing `date_utils` helpers (naive local-time timestamps), so the normalized date matches what `--date-range` would produce for the same input.

### DIFF
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -2,7 +2,7 @@ import os
 
 import typer
 from typing import Optional, List
-from datetime import datetime
+from datetime import datetime, timedelta
 from dateutil import parser as date_parser
 
 # Supported tags for session categorization
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -2,6 +2,7 @@ import os
 
 import typer
 from typing import Optional, List
+from datetime import datetime
 from dateutil import parser as date_parser
 
 # Supported tags for session categorization
@@ -12,7 +13,7 @@ from termstory.parser import parse_all_histories
 from termstory.session import create_sessions
 from termstory.project import detect_projects
 from termstory.database import Database
-from termstory.date_utils import get_current_time, get_today_range
+from termstory.date_utils import get_current_time, get_today_range, parse_date_range_helper
 from termstory.formatter import format_search_results, format_today_output, format_project_output, format_insights_output, format_stats_output, format_profile_output, format_necromancer_score, format_rage_quit_signatures
 import sqlite3
 
@@ -1424,7 +1425,7 @@ def version_callback(value: bool):
 @app.callback(invoke_without_command=True)
 def main(
     ctx: typer.Context,
-    date: Optional[str] = typer.Option(None, "--date", help="Date override (YYYY-MM-DD) for commands"),
+    date: Optional[str] = typer.Option(None, "--date", help="Date override (e.g. today, yesterday, 7days, YYYY-MM-DD) for commands"),
     reset: bool = typer.Option(False, "--reset", help="Reset all TermStory state, configuration, and database"),
     version: Optional[bool] = typer.Option(
         None,
@@ -1441,8 +1442,9 @@ def main(
 
     if date:
         try:
-            date_parser.parse(date)
-            os.environ["TERMSTORY_DATE_OVERRIDE"] = date
+            start_ts, _ = parse_date_range_helper(date)
+            parsed_date = datetime.fromtimestamp(start_ts).strftime("%Y-%m-%d")
+            os.environ["TERMSTORY_DATE_OVERRIDE"] = parsed_date
         except Exception:
             Console(stderr=True).print(f"[bold red]Error: Invalid date format '{date}'[/]")
             raise typer.Exit(code=1)
--- a/tests/test_cli_commands.py
+++ b/tests/test_cli_commands.py
@@ -439,7 +439,23 @@ def test_cli_error_states(tmp_path, monkeypatch):
     result = runner.invoke(app, ["--date", "not-a-date"])
     assert result.exit_code == 1
     assert "Invalid date format" in result.stdout or "Invalid date format" in result.stderr
+def test_global_date_accepts_natural_language(tmp_path, monkeypatch):
+    db_file = tmp_path / "test_date_override.db"
 
+    monkeypatch.setattr("termstory.cli.get_db_path", lambda: str(db_file))
+    monkeypatch.setattr("termstory.config.get_db_path", lambda: str(db_file))
+    monkeypatch.setattr("termstory.cli.get_history_files", lambda: [])
+    monkeypatch.setattr("termstory.cli.run_ingestion", lambda db: None)
+    monkeypatch.setattr("termstory.cli.show_ui", lambda: None)
+
+    runner = CliRunner()
+
+    result = runner.invoke(app, ["--date", "yesterday"])
+
+    assert result.exit_code == 0
+    # The CLI sets TERMSTORY_DATE_OVERRIDE directly, so clean it up
+    # to avoid leaking state into later tests.
+    os.environ.pop("TERMSTORY_DATE_OVERRIDE", None)
 def test_safe_init_db_corrupted(tmp_path, monkeypatch, capsys):
     import sqlite3
     db_file = tmp_path / "corrupt.db"
--- a/tests/test_cli_commands.py
+++ b/tests/test_cli_commands.py
@@ -1,6 +1,6 @@
 import os
 import time
-from datetime import datetime
+from datetime import datetime, timedelta
 from typer.testing import CliRunner
 from termstory.cli import app
 from termstory.database import Database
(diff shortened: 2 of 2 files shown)


Ayush-0918:
:miku /ecsoc

bitflicker64:
:miku /ask why are the tests failing can u give me explaianation and better code snippits for before and after

bitflicker64:
@Ayush-0918 can u fix this

bitflicker64:
@Ayush-0918 wait checking

bitflicker64:
CI is failing on `tests/test_database.py::test_insert_and_retrieve` — reproduced locally.

**Root cause (not in your date parsing):** your new `test_global_date_accepts_natural_language` calls `runner.invoke(app, ["--date", "yesterday"])`, which causes `cli.py` to set `os.environ["TERMSTORY_DATE_OVERRIDE"] = "2026-07-12"` directly. The test never cleans it up.

Because pytest runs files alphabetically, `test_cli_commands` runs before `test_database`. `test_insert_and_retrieve` inserts a session stamped with real `int(time.time())` (today), then calls `get_today_sessions()` → `get_today_range()` → `get_current_time()`. That function honors `TERMSTORY_DATE_OVERRIDE`, so "today" is resolved as yesterday; the real-today session is excluded → `len(today_sessions) == 0`.

**Why `monkeypatch.delenv` after the invoke doesn't help:** monkeypatch records the env state at the moment `delenv`/`setenv` is called. Since the CLI mutates `os.environ` directly inside the invoke, a post-invoke `monkeypatch.delenv` only restores to whatever was current when called (the leaked value). The fix is to purge the variable directly after asserting.

**Fix:** add this at the end of `test_global_date_accepts_natural_language`:

```python
    # The CLI sets os.environ["TERMSTORY_DATE_OVERRIDE"] directly, bypassing
    # monkeypatch. Clean it up manually so later tests see the real "today".
    os.environ.pop("TERMSTORY_DATE_OVERRIDE", None)
```

Verified locally:
- `pytest tests/test_cli_commands.py tests/test_database.py::test_insert_and_retrieve` fails without the cleanup
- Passes with the cleanup

Your actual CLI change for natural-language `--date` looks fine; this is purely test isolation.

```suggestion
    assert result.exit_code == 0

    # The CLI sets os.environ["TERMSTORY_DATE_OVERRIDE"] directly, bypassing
    # monkeypatch. Clean it up manually so later tests see the real "today".
    os.environ.pop("TERMSTORY_DATE_OVERRIDE", None)
```

Ayush-0918:
Thanks! I applied the suggested cleanup using os.environ.pop("TERMSTORY_DATE_OVERRIDE", None) to avoid leaking the environment variable across tests. I verified locally that:

pytest tests/test_cli_commands.py tests/test_database.py::test_insert_and_retrieve -v

termstory/cli.py
```diff
@@ -2,6 +2,7 @@
 
 import typer
 from typing import Optional, List
+from datetime import datetime
 from dateutil import parser as date_parser
```
bitflicker64, line 6:
```suggestion
```
This import is now unused after switching `--date` to `parse_date_range_helper`. Worth removing to keep the diff tidy.

termstory/cli.py
```diff
@@ -1441,8 +1442,9 @@ def main(
 
     if date:
         try:
-            date_parser.parse(date)
-            os.environ["TERMSTORY_DATE_OVERRIDE"] = date
+            start_ts, _ = parse_date_range_helper(date)
+            parsed_date = datetime.fromtimestamp(start_ts).strftime("%Y-%m-%d")
+            os.environ["TERMSTORY_DATE_OVERRIDE"] = parsed_date
```
bitflicker64, line 1447: One subtle thing: `datetime.fromtimestamp(start_ts)` uses the local system timezone. `parse_date_range_helper` returns timestamps built from `datetime.combine(..., time.min)` and `get_current_time()`, which are naive (no tz). Around midnight, a CI runner in a different TZ could shift the override date by one day. For now it works, but if you see flaky date tests later, normalize with `datetime.fromtimestamp(start_ts, tz=timezone.utc).astimezone().date()` or document the behavior.

tests/test_cli_commands.py
```diff
@@ -439,7 +439,23 @@ def test_cli_error_states(tmp_path, monkeypatch):
...
+    monkeypatch.setattr("termstory.config.get_db_path", lambda: str(db_file))
+    monkeypatch.setattr("termstory.cli.get_history_files", lambda: [])
+    monkeypatch.setattr("termstory.cli.run_ingestion", lambda db: None)
+    monkeypatch.setattr("termstory.cli.show_ui", lambda: None)
+
+    runner = CliRunner()
+
+    result = runner.invoke(app, ["--date", "yesterday"])
+
+    assert result.exit_code == 0
```
bitflicker64, line 455:
```suggestion
    assert result.exit_code == 0
    expected = (datetime.now() - timedelta(days=1)).strftime('%Y-%m-%d')
    assert os.environ.get('TERMSTORY_DATE_OVERRIDE') == expected, (
        f"Expected override {expected!r}, got {os.environ.get('TERMSTORY_DATE_OVERRIDE')!r}"
    )
```
Strengthens the test to actually verify the env var value, not just that the CLI didn't crash. Also catches the leak regression if someone removes the cleanup below.

bitflicker64 (commented):
Left a few inline notes. Nothing blocking, mostly polish + one timezone heads-up.

tests/test_cli_commands.py
```diff
@@ -439,7 +439,23 @@ def test_cli_error_states(tmp_path, monkeypatch):
...
+    # The CLI sets TERMSTORY_DATE_OVERRIDE directly, so clean it up
+    # to avoid leaking state into later tests.
+    os.environ.pop("TERMSTORY_DATE_OVERRIDE", None)
 def test_safe_init_db_corrupted(tmp_path, monkeypatch, capsys):
```
bitflicker64, line 463: Good cleanup. For future tests that mutate `os.environ` through the CLI, keep doing this direct `pop` — `monkeypatch.delenv` after the invoke won't catch changes made by the code under test.

bitflicker64:
added good-backend so u get maximum points thanks again !!

Ayush-0918:
Addressed the review feedback.

Changes made:
- Restored the required `date_parser` import used by the search command.
- Added an assertion to verify the normalized `TERMSTORY_DATE_OVERRIDE` value.
- Kept the explicit environment cleanup to prevent state leakage across tests.

Verified locally:

pytest tests/test_cli_commands.py tests/test_database.py::test_insert_and_retrieve -v

Result: 22/22 tests passed.

bitflicker64:
:miku /lgtm 
/approve

bitflicker64 (approved):

PR #282, opened 2026-07-14
feat(ui): add rpg class assigner metric (fixes #38)

### LINKED ISSUE #38: Daily RPG Class & Archetype Assigner 🧙‍♂️

Diwakar-odds:
## Summary
Adds a daily RPG class metric that analyzes command frequency to assign developer archetypes (e.g., "Level 12 Docker Demolitionist") for TUI dashboard gamification. This replaces the previous dictionary-based `assign_rpg_class` with a simpler string-returning `assign_daily_rpg_class` function and integrates the metric into Wrapped and Daily Chronicle view headers.

## Changes
- **Core (`termstory/insights.py`)**:
  - Added `assign_daily_rpg_class()` function that returns a formatted string (e.g., "Level 12 Docker Demolitionist") instead of a dictionary
  - Updated `analyze_all()` to call the new function and extract the class string directly
- **TUI (`termstory/tui.py`)**:
  - Updated imports and calls in `calculate_dashboard_stats()` to use `assign_daily_rpg_class`
  - Added "DAILY CLASS" metric to Wrapped view header in `_render_wrapped_view_ui()`
  - Added "DAILY CLASS" metric to Daily Chronicle view header in `render_daily_chronicle_view()`
  - Added traceback import in `refresh_details_canvas()` for debugging
- **Tests (`tests/test_insights.py`)**:
  - Added `test_assign_daily_rpg_class()` covering empty sessions, known commands (kubectl), and fallback for unknown commands

## Architecture / Flow
```mermaid
flowchart LR
    A[Session Commands] --> B[assign_daily_rpg_class]
    B --> C[Command Frequency Count]
    C --> D[Class Map Lookup]
    D --> E[Level Calculation]
    E --> F[Formatted String Output]
    F --> G[TUI Headers]
    F --> H[analyze_all]
```

## Fixes
- Fixes #38 — Implements RPG Class Assigner metric for user behavior analysis

## Testing
Run `pytest tests/test_insights.py::test_assign_daily_rpg_class` to verify the new function handles empty sessions, known commands, and fallback scenarios. To test the TUI integration, run `termstory gui` and navigate to any Wrapped or Daily Chronicle view to confirm the "DAILY CLASS" metric renders correctly in the header.

## Notes
The new `assign_daily_rpg_class()` function returns a string directly, unlike the previous `assign_rpg_class()` which returned a dictionary with a "class_name" key. This is an internal API change but maintains backward compatibility in the `analyze_all()` return value. The function includes special handling for piped commands (defaults to "grep" for Regex Sorcerer logic) and strips basename from command paths using `os.path.basename()`.


Diwakar-odds:

This PR introduces the RPG Class Assigner (Issue #38) to the TUI dashboards. It calculates the user's primary command usage and dynamically assigns them an RPG developer archetype (e.g. "Docker Demolitionist").

- `ECSoC26-L2` as this introduces significant UI features and analytics rendering.
- `good-ui` and `good-issue` as this implements highly requested visual gamification elements and complex data grouping.

bitflicker64 (dismissed):
@/tmp/pr282-review-body.md

bitflicker64 (dismissed):
test...

bitflicker64 (commented):
test

bitflicker64 (commented):
test

bitflicker64 (commented):
test

bitflicker64 (commented):
test

termstory/insights.py
```diff
@@ -206,6 +206,47 @@ def calculate_streak(sessions: List[Session]) -> int:
             break
     return streak
 
+def assign_rpg_class(sessions: List[Session]) -> str:
```
bitflicker64: This is the duplicate function. There's already an `assign_rpg_class(sessions) -> Dict[str, Any]` further down in this same file (around line 548 on the PR head). In Python the last `def` at module scope wins, so `from termstory.insights import assign_rpg_class` resolves to the dict version, not this new string version. Rename this one (e.g. `assign_daily_rpg_class`) and update the three call sites in `termstory/tui.py` to use the new name, or alternatively delete the old dict version and migrate its callers (`tui.py:183-195` and `tests/test_insights.py:176-201`).

termstory/tui.py
```diff
@@ -929,7 +931,7 @@ def render_time_summary(
         header_lines.append(f"[bold cyan]{avatar_lines[6]}[/]     [bold cyan]ACTIVE REPOS:[/]      [bold]{active_projects_count} Workspaces[/]")
         header_lines.append(f"[bold cyan]{avatar_lines[7]}[/]     [bold cyan]FOCUS SCORE:[/]     [bold green]{fs:.1f}/10.0[/]")
         header_lines.append(f"[bold cyan]{avatar_lines[8]}[/]     [bold cyan]PEAK VELOCITY:[/]    [dim]{peak_velocity}[/]")
-        header_lines.append(f"[bold cyan]{avatar_lines[9]}[/]     [bold cyan]PROJECTS:[/]        [dim]{active_projects_count}[/]")
+        header_lines.append(f"[bold cyan]{avatar_lines[9]}[/]     [bold cyan]DAILY CLASS:[/]      [dim]{rpg_class}[/dim][/]")
```
bitflicker64: Root cause of the CI failure (`MarkupError: auto closing tag ('[/]') has nothing to close`). Because of the duplicate `assign_rpg_class` at line 548, `rpg_class` here is actually a dict, not a string. The f-string interpolates the dict repr (`{'class_name': 'Regex Sorcerer', ...}`) into a Rich markup string, and the `[` in the repr breaks the parser. Fix the duplicate upstream and this resolves.

termstory/tui.py
```diff
@@ -1160,7 +1164,7 @@ def _render_wrapped_view_ui(
         header_lines.append(f"[bold cyan]{avatar_lines[6]}[/]     [bold cyan]ACTIVE REPOS:[/]      [bold]{active_projects_count} Workspaces[/]")
         header_lines.append(f"[bold cyan]{avatar_lines[7]}[/]     [bold cyan]FOCUS SCORE:[/]     [bold green]{fs:.1f}/10.0[/]")
         header_lines.append(f"[bold cyan]{avatar_lines[8]}[/]     [bold cyan]PEAK VELOCITY:[/]    [dim]{peak_velocity}[/]")
-        header_lines.append(f"[bold cyan]{avatar_lines[9]}[/]     [bold cyan]COMMITS:[/]         [dim]{total_commits}[/]")
+        header_lines.append(f"[bold cyan]{avatar_lines[9]}[/]     [bold cyan]DAILY CLASS:[/]      [dim]{rpg_class}[/dim][/]")
```
bitflicker64: Same `MarkupError` root cause as the timeframe header. Once the duplicate `assign_rpg_class` is resolved, this line is fine.

termstory/tui.py
```diff
@@ -1357,7 +1363,7 @@ def render_daily_chronicle_view(self, date_str: str, sessions: List[Session], pr
         header_lines.append(f"[bold cyan]{avatar_lines[6]}[/]     [bold cyan]ACTIVE SESSIONS:[/] [bold]{len(sessions)}[/]")
         header_lines.append(f"[bold cyan]{avatar_lines[7]}[/]     [bold cyan]FOCUS SCORE:[/]     [bold green]{fs:.1f}/10.0[/]")
         header_lines.append(f"[bold cyan]{avatar_lines[8]}[/]     [bold cyan]PEAK TIME:[/]       [dim]{peak_velocity}[/]")
-        header_lines.append(f"[bold cyan]{avatar_lines[9]}[/]     [bold cyan]PROJECTS:[/]        [dim]{len(projects)}[/]")
+        header_lines.append(f"[bold cyan]{avatar_lines[9]}[/]     [bold cyan]DAILY CLASS:[/]      [dim]{rpg_class}[/dim][/]")
```
bitflicker64: Same `MarkupError` root cause. Fix the duplicate upstream.

termstory/tui.py
bitflicker64: Wrapping the `COMMITS:` row instead of replacing it keeps the commit-volume metric visible in monthly wrapped summaries. `total_commits` is still computed above, just no longer rendered.

bitflicker64 (changes requested):
Hey @Diwakar-odds, thanks for picking up #38. The idea is fun and the helper logic is mostly sound, but I hit a real P1 while reading and CI is red on all four Python versions, so this needs another pass before merge.

## What's blocking

1. **Duplicate `assign_rpg_class` in `termstory/insights.py`.** Your new string-returning version lives at line 209. There's already a dict-returning `assign_rpg_class` at line 548 that the codebase depends on. In Python the last `def` wins, so every `from termstory.insights import assign_rpg_class` resolves to the dict version. That means your new helper is dead code on import, and the three header call sites in `tui.py` are actually receiving a dict, not a string.

2. **That's why CI is red.** `MarkupError: auto closing tag ('[/]') has nothing to close` in `test_tui_landing_page_after_onboarding` comes from the new `DAILY CLASS` header interpolating a dict repr (`{'class_name': 'Regex Sorcerer', ...}`) into a Rich markup string. The unescaped `[` from the dict breaks the parser. Same pattern at all three header sites.

3. **Wrapped view loses the commit count.** In `_render_wrapped_view_ui` you replaced `[bold cyan]COMMITS:[/] [dim]{total_commits}[/]` with the new class line, but `total_commits` is still computed above. Monthly wrapped summaries no longer surface commit volume.

## Suggested fix shape

- Rename your new helper (e.g. `assign_daily_rpg_class`) and remove the dead duplicate at line 209, **or** delete the old dict-returning one and migrate its two callers (`tui.py:183-195` and `tests/test_insights.py:176-201`). Renaming is the lower-risk path.
- Restore the `COMMITS:` row in `_render_wrapped_view_ui` instead of replacing it. You have header lines to spare.
- Add a test for the new helper (empty sessions, no commands, single command, top-cmd fallback to `Scripting Shaman (foo)`).

## Smaller notes

- `level = min(100, max(1, total_cmds // 50))` means 50 commands = Level 1, 100 = Level 2. The issue's example is "Level 12 Regex Sorcerer", which is 600 commands. Either drop the threshold or normalize against the dominant-class share rather than raw totals, or the level will feel arbitrary.
- `cmd.split()[0]` won't match a top command if shell history stores it as a path (e.g. `~/.local/bin/python`). Strip the basename.
- `class_map` doesn't include `python3`, `pip`, `pytest`, `npx`, `pnpm`, `yarn`, `gh`, `make`, `awk`, `sed`. The old dict-returning version covers more ground via `lower_cmd.split()` substring match. Worth keeping that coverage in the new map.
- Trailing whitespace on a few blank lines inside the new function (PEP-8 nit, not blocking).

Happy to approve once the duplicate is resolved and CI is green.

bitflicker64:
Quick housekeeping note: there are 4 stray "test" review stubs above from when I was probing the API to confirm which diff paths GitHub accepts for inline comments. I couldn't delete them after the fact (GitHub only allows deleting pending reviews). Sorry for the noise. The real review with inline comments is the **CHANGES_REQUESTED** one directly above. — bitflicker

Diwakar-odds:
I've resolved the duplicate ssign_rpg_class function (renamed the new one to ssign_daily_rpg_class), restored the COMMITS line in the wrapped view, added tests for the new helper, fixed the thresholds, included the extra commands in the class map, and resolved the race conditions that were causing CI to fail. CI should be fully green now! Please review.

termstory/tui.py
```diff
@@ -893,12 +893,14 @@ def render_time_summary(
         
         active_project_ids = {s.project_id for s in sessions}
         active_projects_count = len(active_project_ids)
-        total_commits = sum(len(s.commits) for s in sessions)
+        total_commits = sum(len(s.commands) for s in sessions)
```
bitflicker64, line 906: This is now counting commands while the label still says `COMMITS`. The same change appears in the wrapped and daily chronicle headers too, so users will see command volume reported as commit volume. Please keep this as `sum(len(s.commits) for s in sessions)` or rename the label if the intent is to show commands.

termstory/tui.py
```diff
@@ -1074,6 +1076,8 @@ def render_time_summary(
                         feed_widgets.append(Static(f"  └─ {escape(heur)}\n"))
                         
             self.mount(Vertical(*feed_widgets, classes="feed-container"))
```
bitflicker64, line 1085: This debug print will write to stderr every time the feed container mounts in normal TUI use. Please remove it or route it through `logger.debug(...)`; otherwise launching/navigating the dashboard leaves noisy terminal output.

termstory/tui.py
```diff
@@ -2321,6 +2329,8 @@ def refresh_details_canvas(self) -> None:
             
             self._show_node_details(selected_node)
         except Exception as e:
```
bitflicker64, line 2333: This path intentionally suppresses UI exceptions, but `traceback.print_exc()` now dumps a full stack trace to stderr for recoverable UI errors. Please remove it or keep the existing `logger.debug(...)` path so normal TUI usage stays quiet.

bitflicker64 (changes requested):
Two regressions remain on the current green-CI head: the COMMIT(S) header is counting commands, and debug output was left in normal TUI paths.

bitflicker64:
:miku /label good-backend

bitflicker64:
Get this merch, and you'll get the maximum points you can from me.

bitflicker64:
Can you confirm if these labels are stackable? If yes, I can stack them too.

bitflicker64:
:miku /check-star

bitflicker64 (approved):

Diwakar-odds:
Hi @bitflicker64, I noticed this PR was merged/closed but seems to be missing the `good pr` label. Could you please take a look and add the label if it qualifies? Thank you!

PR #289, opened 2026-07-17
test: add regression test for TimestampDetective when project paths callback fails

### LINKED ISSUE #286: Add regression test for TimestampDetective project path callback failures
## Summary
Add a regression test in `tests/test_timestamp_detective.py` to verify that the TimestampDetective integration gracefully handles project path callback failures during history parsing.

## Steps to Reproduce (if bug)
Omitted - this is a feature request.

## Expected vs Actual
**Expected**: When the `project_paths` callback raises an exception during `parse_zsh_history`, the parser should catch the exception, log it, and continue ingestion with an empty project path list. The TimestampDetective should still process legacy commands and return usable command/session data.

**Actual**: There is currently no focused test in `tests/test_timestamp_detective.py` that documents this degraded behavior. A test exists at the parser level (`test_parse_zsh_history_project_paths_callback_raises` in `tests/test_parser.py`), but a dedicated test for the TimestampDetective integration is missing.

## Environment
Not specified.

## Context
The `parse_zsh_history` function accepts a `project_paths` parameter that can be either a list of strings or a callable returning a list of strings . When the callable version raises an exception, the current implementation catches it broadly and continues with an empty project list . This ensures that project path enrichment is treated as a nice-to-have feature rather than a core parsing requirement. The existing parser-level test verifies this behavior , but a focused test in the TimestampDetective test suite would better document the intended degraded behavior at the integration layer.

deepmhatre13:
Closes #286 
### Summary

Adds a regression test covering the scenario where the parser's project paths callback raises an exception and falls back to an empty `project_paths` list before creating `TimestampDetective`.

### What this tests

- Simulates the parser fallback by initializing `TimestampDetective` with `project_paths=[]`.
- Verifies that `resolve_all()` does not raise an exception.
- Confirms that all legacy commands are returned without being dropped.
- Ensures each returned item still contains the expected timestamp-related fields:
  - `detected_ts`
  - `detected_source`
  - `is_legacy_still`

### Why

This prevents regressions where an exception during project path resolution could cause `TimestampDetective` to fail or lose legacy commands. The test verifies that the fallback behavior continues to function correctly even when no project paths are available.

### Validation

- ✅ Added regression test:
  - `TestResolveAll::test_project_paths_callback_raises_still_resolves_legacy`
- ✅ Regression test passes successfully.
- ✅ Existing parser callback regression test also passes.
- ✅ No production code changes.

### DIFF
--- a/tests/test_timestamp_detective.py
+++ b/tests/test_timestamp_detective.py
@@ -17,12 +17,10 @@ Covers:
 
 import os
 import sys
-import glob
-import stat
 import time
 import tempfile
 import unittest
-from unittest.mock import patch, MagicMock, call
+from unittest.mock import patch, MagicMock
 from datetime import datetime, timezone, timedelta
 
 # Ensure the project root is on the path
@@ -34,6 +32,7 @@ from termstory.timestamp_detective import TimestampDetective
 # Helpers
 # ---------------------------------------------------------------------------
 
+
 class ParserBugDateTime:
     @staticmethod
     def strptime(*_args, **_kwargs):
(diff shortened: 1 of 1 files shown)


deepmhatre13:
:miku /ecsoc

deepmhatre13:
@bitflicker64 ,kindly review the pr

bitflicker64 (approved):
Looks good to me. Thanks for the contribution.

PR #290, opened 2026-07-17
fix: prevent onboarding modal freeze on Textual 8.x

### LINKED ISSUE #288: Onboarding Save & Enable freezes with current Textual 8.x installs
## What happens

Fresh installs from PyPI or `scripts/install.sh` can get stuck on the first-run TermStory onboarding screen.

A user can enter their API key, press **Save & Enable**, and then nothing visibly happens: the modal stays open, the app feels frozen, and the user never gets dropped into the actual dashboard. This makes the first-run experience feel broken right at the moment where someone has trusted the app enough to paste in an API key.

## Why this matters

This affects the normal install path, not just local development:

```bash
pip install termstory
termstory ui
```

or:

```bash
curl -fsSL https://raw.githubusercontent.com/bitflicker64/Termstory/main/scripts/install.sh | bash
termstory ui
```

`pyproject.toml` currently allows any Textual version newer than `0.50.0`:

```toml
textual>=0.50.0
```

Today, that resolves to Textual `8.x`, and the onboarding dismiss path in `termstory/tui.py` is not compatible with that version.

## Reproduction

On a fresh/main install with Textual `8.2.x`:

1. Remove or reset existing TermStory config so onboarding appears.
2. Run `termstory ui`.
3. Enter a Groq/OpenAI API key in the onboarding modal.
4. Click **Save & Enable**.
5. The modal does not close and the UI appears frozen.

## Root cause

The onboarding save path schedules modal dismissal through a zero-delay Textual timer:

```python
self.set_timer(0.0, lambda: self.dismiss(self.config))
```

There are several similar calls in `termstory/tui.py`, including Save/Disable/Escape flows for modal screens.

With Textual `8.2.7`, a minimal reproduction against `main` triggers this internal failure:

```text
Task exception was never retrieved
future: <Task finished name='set_timer#1' ... exception=ZeroDivisionError('float division by zero')>
Traceback (most recent call last):
  File ".../textual/timer.py", line 163, in _run
    count = int((now - start) / _interval + 1)
ZeroDivisionError: float division by zero
```

Because the timer crashes before the callback completes, `ModalScreen.dismiss()` never delivers the onboarding result back to `TermStoryWorkspace.handle_onboarding_result()`. That means the config is not saved and the screen never exits.

## Extra clue from the repo

The tests already hint at this family of issues. `tests/test_tui.py` avoids actually pressing `#btn-save` because of Textual 8.x modal dismiss behavior, and `termstory/testing.py` documents a workaround for `ModalScreen.dismiss()` / `AwaitComplete` hangs. So the production path needs the same level of care, not just the tests.

## Suggested fix

Avoid zero-delay timers for modal dismissal and avoid returning Textual's awaitable dismiss object from the timer callback.

A minimal shape that fixed the local reproduction was:

```python
self.set_timer(0.001, lambda: (self.dismiss(self.config), None)[1])
```

For consistency, apply the same pattern or a small helper to all modal-dismiss timer sites, for example:

```python
def dismiss_later(self, result=None) -> None:
    self.set_timer(0.001, lambda: (self.dismiss(result), None)[1])
```

Then replace paths like:

```python
self.set_timer(0.0, lambda: self.dismiss(self.config))
self.set_timer(0.0, lambda: self.dismiss(None))
self.set_timer(0.0, self.dismiss)
```

with the helper.

## Acceptance criteria

- Pressing **Save & Enable** closes the onboarding modal and persists config.
- Pressing **Keep Local Only** also closes the modal and persists `has_seen_onboarding`.
- Escape/cancel modal paths do not freeze or emit timer exceptions.
- Add a regression test that actually presses `#btn-save` under the installed Textual version instead of bypassing the button path.
- Optional but useful: tighten the Textual dependency range if the app only supports specific major versions.

## Environment observed

- TermStory `main` / packaged install path
- Textual `8.2.7` from current pip resolution
- Failure reproduced with a fresh checkout of `bitflicker64/Termstory` main

This is one of those bugs that feels small in code but huge in user experience: the first-run screen becomes a dead end right after the user enters their API key.

deepmhatre13:
Closes #288 
### Summary

This PR fixes the onboarding modal freeze that occurs on Textual 8.x when clicking **Save & Enable**.

The root cause was the use of `set_timer(0.0, ...)` for modal dismissal, which can trigger a `ZeroDivisionError` in newer Textual versions and prevent the onboarding flow from completing.

### Changes

- Added a reusable `dismiss_later()` helper for modal screens.
- Replaced all `set_timer(0.0, ...)` modal dismissal calls with `dismiss_later()`.
- Updated the following modal screens:
  - `HelpScreen`
  - `OnboardingScreen`
  - `ResetConfirmScreen`
  - `MatrixDefragScreen`
  - `GhostTyperScreen`
- Used a small non-zero delay (`0.001s`) to avoid the Textual timer issue.
- Ensured the scheduled callback returns `None`, preventing `AwaitComplete`/`ScreenError` issues during modal dismissal.

### Regression Test

Added an end-to-end regression test for the onboarding flow.

The test:

- launches the application with onboarding enabled,
- fills the API key field using a fake non-empty key (`gsk_test_key`),
- triggers the actual **Save & Enable** button logic,
- verifies the onboarding modal closes successfully,
- confirms the onboarding result is applied to the application configuration.

The test uses `widget.press()` because `pilot.click("#btn-save")` results in an `OutOfBounds` error in the test environment when the button is outside the visible viewport. `widget.press()` dispatches the same `Button.Pressed` event and exercises the production code path:

```
on_button_pressed()
    ↓
dismiss_later()
    ↓
handle_onboarding_result()
```

### Validation

Passed:

```bash
pytest tests/test_tui.py -k onboarding -v
```

Result:

```
5 passed
```

Passed:

```bash
pytest tests/test_tui.py -k "onboarding or api_key_validation" -v
```

Result:

```
6 passed
```

Verified that:

- the onboarding modal no longer freezes,
- no `ZeroDivisionError` occurs,
- no `ScreenError` is raised,
- onboarding configuration is saved successfully.

### Notes

The remaining failures in the full `tests/test_tui.py` suite are pre-existing `NoMatches` failures related to `DetailsCanvas` rendering (`#btn-exec-*`, `#btn-bulk-*`, `.feed-container`) and are unrelated to this onboarding modal fix.


deepmhatre13:
:miku

bitflicker64:
@deepmhatre13 Can you prove the fix works? Can you record a video and send it here? I'll definitely give you a good backend label so you get the maximum points in ESOC. Also, if you find any other issues, can you just file them too? If you want, you can claim them. I'll definitely give you good labels so you can score better in ECSOC.

bitflicker64:
I don’t have much time to search for and file issues here. I have a reliable bot that can file them, but it’s often just the usual situation. I was wondering if you could file some issues and perhaps claim them, or at least allow me to add the ECSOC labels to them. Also, the good labels can earn extra points, definitely.

bitflicker64:
Also, could you find out whether the labels are stackable? If they are, I can give you multiple labels to earn a lot of points. I would appreciate it if you could look into this. Thank you.

deepmhatre13:
Hi @bitflicker64,

I've verified the fix locally and in CI.

**Local verification**

* Ran the onboarding regression suite:

  ```bash
  pytest tests/test_tui.py -k "onboarding or api_key_validation" -v
  ```
* Result: **6 passed, 25 deselected**.

This includes the new end-to-end onboarding regression test that exercises the actual **Save & Enable** flow.

**CI verification**

* All GitHub checks passed:

  * Python 3.9 ✅
  * Python 3.10 ✅
  * Python 3.11 ✅
  * Python 3.12 ✅
  * Lint ✅
  * Greptile ✅

I've attached screenshots of the local test results and the successful CI checks as proof.
image
image

deepmhatre13:
Hi @bitflicker64! I checked the publicly available ECSoC documentation, but I couldn't find any official mention of whether labels are stackable for scoring. It doesn't seem to be covered in the program rules.

If there's an internal guideline or admin clarification about label stacking, I'd be happy to follow that. Otherwise, I couldn't verify that multiple labels contribute additional points.

bitflicker64:
> Hi @bitflicker64! I checked the publicly available ECSoC documentation, but I couldn't find any official mention of whether labels are stackable for scoring. It doesn't seem to be covered in the program rules.
> 
> If there's an internal guideline or admin clarification about label stacking, I'd be happy to follow that. Otherwise, I couldn't verify that multiple labels contribute additional points.


https://github.com/user-attachments/assets/820c8a83-c91f-4f4b-9836-9af034acee96

bitflicker64:
@deepmhatre13 I'm currently working on containerization in one of  Apache projects , so I'm a bit busy, but I'm still finding time to test it. I'll quickly make a video. You can test it in WSL since you have a Windows laptop, not a Mac or Linux. This was originally intended for Mac, but it should work on Linux as well. You can try it with WSL. You can contact me here, or we can connect on Discord. I'll send you a link right now; let me know if it works. Please confirm here, and I'll re‑run it. https://discord.gg/PHcXrachS

bitflicker64:
:miku /label good-backend
:miku /label good-pr

bitflicker64:
Until then, you can take on small PRs so you continue earning points, too. Thanks.

bitflicker64:
Thanks for fixing the Textual timer crash. I think there is one more practical issue left: the onboarding modal is still tall enough that **Save & Enable** can sit below the visible area in normal terminals such as `120x40`, so the flow can still look stuck even though the timer crash is fixed.

The fix I would apply is to shorten the onboarding copy and reduce the modal vertical spacing:

```diff
diff --git a/termstory/tui.py b/termstory/tui.py
@@
             Static("🔒 TermStory Privacy & AI Onboarding", id="modal-title"),
             Static(
-                "TermStory is completely offline by default. No data ever leaves your machine.\n\n"
-                "To make your timeline readable, you can optionally enable the AI Categorization Engine. "
-                "This will send sanitized terminal commands to an LLM to generate short, human-readable session memories.\n\n"
+                "TermStory is offline by default. If you enable AI, only sanitized terminal commands are sent "
+                "to your selected provider to create short session memories.\n\n"
                 "Before sending, TermStory scrubs passwords, environment variables, IPs/FQDNs, and drops "
-                "sensitive sessions containing commands like 'vault' or 'aws configure' entirely.\n\n"
-                "[dim]Shortcuts: [bold]Ctrl+g[/]: Groq, [bold]Ctrl+a[/]: OpenAI, [bold]Ctrl+l[/]: Ollama, [bold]Ctrl+c[/]: Custom, [bold]Ctrl+d[/]: Disable AI, [bold]Esc[/]: Cancel[/dim]",
+                "sensitive sessions such as vault or aws configure.\n\n"
+                "[dim]Shortcuts: Ctrl+g Groq, Ctrl+a OpenAI, Ctrl+l Ollama, Ctrl+c Custom, Ctrl+d Disable, Esc Cancel[/dim]",
                 id="modal-desc"
             ),
@@
         width: 90%;
         max-width: 75;
         height: auto;
-        padding: 2;
+        padding: 1;
         content-align: center middle;
@@
     .input-label {
         color: #88889a;
-        margin-top: 1;
+        margin-top: 0;
         margin-left: 1;
         text-style: bold;
     }
```

I checked this locally on the PR branch:

```bash
pytest tests/test_tui.py -k "api_key_validation" -q
# 1 passed, 30 deselected
```

I also probed the Textual layout at `120x40`; after this patch, `#btn-save` renders at `y=35`, height `3`, so it stays within the 40-row screen and remains visible/clickable. This complements the timer fix rather than replacing it.

termstory/tui.py
```diff
@@ -404,15 +404,14 @@ def compose(self) -> ComposeResult:
...
-            self.set_timer(0.0, self.dismiss)
+            self.dismiss_later()
 
     def action_dismiss_none(self) -> None:
-        self.set_timer(0.0, self.dismiss)
+        self.dismiss_later()
+
+    def dismiss_later(self, result=None) -> None:
+        """Dismiss this modal screen on the next tick with a non-zero delay."""
+        self.set_timer(0.001, lambda: (self.dismiss(result), None)[1])
```
bitflicker64: Works, but the tuple trick is hard to read and this method is duplicated on several screens. Consider:

```python
def dismiss_later(self, result=None) -> None:
    """Dismiss on the next timer tick (non-zero delay required on Textual 8.x)."""
    def _do_dismiss() -> None:
        self.dismiss(result)
    self.set_timer(0.001, _do_dismiss)
```

and/or a shared mixin so Help/Onboarding/Reset/Matrix/GhostTyper don’t each redefine it.

tests/test_tui.py
```diff
@@ -1042,61 +1042,66 @@ async def test_tui_deep_search_scope_escape():
...
+            save_btn.press()
+
+            # Wait for the dismissal timer to fire and the screen to pop.
+            for _ in range(50):
+                if not isinstance(app.screen, OnboardingScreen):
+                    break
+                await pilot.pause()
+
+            # Verify the modal dismissed and the result reached the app.
+            assert not isinstance(app.screen, OnboardingScreen)
```
bitflicker64, line 1117: Nice end-to-end path. Please double-check (or monkeypatch) that Save does not persist to the real user config under `$HOME` during CI/dev runs — e.g. `monkeypatch.setattr("termstory.tui.save_config", lambda cfg: None)` if the onboarding success path writes config.

bitflicker64 (commented):
## Code Review Summary

**Verdict: Approve with minor suggestions** — real fix for Textual 8.x modal freeze; good E2E regression test. A few cleanups would make this more maintainable.

### Intent
Stop onboarding / modal dismiss from freezing under Textual 8.x by scheduling dismiss off the button message pump with a non-zero timer delay (issue #288).

### Impact
- Replaces `set_timer(0.0, ...)` with `dismiss_later` using `0.001`s delay and optional result.
- Applied across Help, Onboarding, ResetConfirm, MatrixDefrag, GhostTyper screens.
- Test rewritten to press real Save & Enable path and wait for screen pop — much better than the previous structural mock.

### What’s correct
- Root cause analysis (AwaitComplete / ScreenError on zero-delay dismiss inside message handler) matches known Textual 8 behavior.
- CI green.
- Non-empty API key path exercised without network.

### Suggestions
1. **DRY `dismiss_later`** — identical 3-line method is copy-pasted on 5 classes. Prefer a small mixin / helper base, e.g. `class DeferredDismissMixin: ...`, to avoid future drift.
2. **Lambda style** — `(self.dismiss(result), None)[1]` works but is opaque; `lambda r=result: self.dismiss(r)` (or a tiny local def) is clearer and avoids the tuple trick.
3. **Config isolation in the new test** — confirm `save_config` cannot write the developer’s real config file when Save is pressed (monkeypatch `termstory.tui.save_config` / config path if not already covered by workspace config override).
4. **Comment why `0.001` not `0.0`** — one line in `dismiss_later` docstring helps the next person not “optimize” it back to zero.

### Verdict
Comment / soft approve — mergeable after optional DRY + config-safety check; not blocking if you confirm the test doesn’t touch real HOME config.

bitflicker64:
:miku /check-star

deepmhatre13:
:miku /check-star

PR #304, opened 2026-07-18
[Web] Fix report session truncation

### LINKED ISSUE #300: [Medium] web.py: unfiltered report caps at 30 sessions, filtered at 1000 — silent inconsistent truncation
## Bug Description

HTML report data loading silently truncates sessions with two different hard caps depending on whether a date filter is present:

| Mode | SQL LIMIT | Effect |
|------|-----------|--------|
| No `start_ts`/`end_ts` | **30** | Full “all time” report only shows the 30 newest sessions |
| With date filter | **1000** | Broader window, still silent cap |

Users who open a report without dates think they are seeing their whole history; heavy users only ever see a thin recent slice. The limits are magic numbers with no config and no UI warning.

## Location

`termstory/web.py` — `get_web_data` session id query (~lines 50–65):

```python
if conditions:
    query += " WHERE " + " AND ".join(conditions)
    query += " ORDER BY start_time DESC LIMIT 1000"
else:
    query += " ORDER BY start_time DESC LIMIT 30"
```

Related hardcoded cap: AI extra sessions `LIMIT 15` (~line 173).

## Steps to Reproduce

1. Populate DB with >30 sessions across multiple months.
2. Generate web report with no date filter (`termstory web` / equivalent).
3. Actual: only 30 newest sessions appear in report data.
4. Generate report with a wide date range covering all sessions.
5. Actual: up to 1000 sessions — different subset rules, still silent.

## Expected Behavior

- Document the cap in the report UI (“Showing N of M sessions”) **or** raise the unfiltered default to match the filtered path.
- Prefer a single named constant (e.g. `WEB_REPORT_SESSION_LIMIT`) and/or config key.
- If truncation happens, surface it in the returned data payload so the template can warn.

## Suggested Fix

1. Introduce module-level constants for the limits.
2. Unify defaults (recommend one sensible cap, e.g. 1000, for both paths) unless product intent is deliberately “recent-only” without filter — if so, document it in the report header.
3. Return `truncated: bool` / `total_sessions` in the data dict for the template.

## Acceptance Criteria

- [ ] Limits are named constants (not raw literals only)
- [ ] Unfiltered vs filtered behavior is intentional and documented in-code
- [ ] Report data indicates when results were truncated
- [ ] Test asserts truncation metadata or consistent limit application

floze-the-genius:
## Summary
Unifies session limit logic for web reports under named constants, adds transparent truncation metadata, and displays a visible notice when reports are truncated.

### Core
- Extracted `WEB_REPORT_SESSION_LIMIT = 500` and `AI_HIGHLIGHT_LIMIT = 15` constants in `termstory/web.py` to replace inline magic numbers 
- Removed redundant if/else branch in session query that both applied `LIMIT 500`, replaced with single parameterized query using `WEB_REPORT_SESSION_LIMIT` 
- Parameterized AI highlights query to use `AI_HIGHLIGHT_LIMIT` instead of inline `LIMIT 15` 
- Added `session_window` metadata to returned dict with `limit`, `returned`, `total`, and `truncated` fields 

### UI
- Added CSS for `.session-window-notice` with `.visible` modifier class 
- Added HTML notice element with `role="status"` and `aria-live="polite"` attributes 
- Added JavaScript to display "Showing X of Y sessions." notice when `sessionWindow.truncated` is true 

### Tests
- Added `session_window` assertion to `test_get_web_data_empty_db` 
- Added `test_get_web_data_does_not_truncate_unfiltered_sessions_at_30` to verify 31 sessions return without truncation 
- Updated `test_generate_and_open_report` to verify notice HTML and JavaScript wiring 
- Updated `test_swarm_audit_fixes` to derive overflow fixture from `WEB_REPORT_SESSION_LIMIT` instead of hardcoding 1000/1005 

## Architecture / Flow

```mermaid
flowchart LR
    A[get_web_data] --> B[Query Sessions]
    B --> C[Apply WEB_REPORT_SESSION_LIMIT]
    C --> D[Build session_window Metadata]
    D --> E[Return to Frontend]
    E --> F{truncated?}
    F -->|Yes| G[Show Notice]
    F -->|No| H[Hide Notice]
```

## Fixes
- Fixes #300 — unifies filtered and unfiltered web reports around a named session limit instead of silently restricting unfiltered reports

## Testing
- `python -m pytest tests/test_web.py -q --timeout=60` — 9 passed
- `uvx ruff check termstory/web.py` — passed
- `git diff --check` — passed

## Notes
- Rebased onto `upstream/main@1996a2f` and resolved conflict by retaining upstream's unified 500-session cap through `WEB_REPORT_SESSION_LIMIT` while preserving the PR's parameterized query, `session_window` metadata, truncation notice, and regression coverage.

### DIFF
--- a/termstory/web.py
+++ b/termstory/web.py
@@ -480,6 +480,18 @@ def generate_and_open_report(
             color: var(--text-secondary);
             font-size: 0.9rem;
         }}
+        .session-window-notice {{
+            display: none;
+            margin: -16px 0 28px;
+            padding: 12px 16px;
+            border-left: 3px solid var(--accent-color);
+            background: var(--panel-bg);
+            color: var(--text-secondary);
+            font-size: 0.9rem;
+        }}
+        .session-window-notice.visible {{
+            display: block;
+        }}
 
         /* Glassmorphism panel */
         .panel {{
@@ -851,6 +863,8 @@ def generate_and_open_report(
                 <div style="margin-top: 2px;">Last Ingested: <span id="last-ingested-date">-</span></div>
             </div>
         </header>
+
+        <div id="session-window-notice" class="session-window-notice" role="status" aria-live="polite"></div>
         
         <div class="kpi-grid">
             <div class="panel kpi-card">
@@ -1179,6 +1193,13 @@ def generate_and_open_report(
             document.getElementById('kpi-commands').textContent = reportData.stats.total_commands;
             document.getElementById('kpi-projects').textContent = reportData.stats.total_projects;
             document.getElementById('kpi-streak').textContent = `${{reportData.stats.streak}} Days`;
+
+            const sessionWindow = reportData.session_window;
+            if (sessionWindow?.truncated) {{
+                const notice = document.getElementById('session-window-notice');
+                notice.textContent = `Showing ${{sessionWindow.returned}} of ${{sessionWindow.total}} sessions.`;
+                notice.classList.add('visible');
+            }}
             
             const searchInput = document.getElementById('search-bar');
             searchInput.addEventListener('input', (e) => {{
--- a/termstory/web.py
+++ b/termstory/web.py
@@ -10,6 +10,9 @@ from termstory.database import Database
 
 logger = logging.getLogger(__name__)
 
+WEB_REPORT_SESSION_LIMIT = 500
+AI_HIGHLIGHT_LIMIT = 15
+
 
 def get_web_data(db: Database, start_ts: Optional[int] = None, end_ts: Optional[int] = None) -> dict:
     """Gather stats, project list, timeline, and AI highlights from the database, filtering by date range if provided."""
@@ -58,10 +61,9 @@ def get_web_data(db: Database, start_ts: Optional[int] = None, end_ts: Optional[
             
         if conditions:
             query += " WHERE " + " AND ".join(conditions)
-            query += " ORDER BY start_time DESC LIMIT 500"
-        else:
-            query += " ORDER BY start_time DESC LIMIT 500"
-            
+        query += " ORDER BY start_time DESC LIMIT ?"
+        params.append(WEB_REPORT_SESSION_LIMIT)
+
         cursor.execute(query, params)
         session_ids = [row[0] for row in cursor.fetchall()]
     finally:
@@ -152,7 +154,7 @@ def get_web_data(db: Database, start_ts: Optional[int] = None, end_ts: Optional[
 
     # 4. AI Summary Highlights
     ai_sessions = [s for s in sessions_data if s["ai_summary"]]
-    if len(ai_sessions) < 15:
+    if len(ai_sessions) < AI_HIGHLIGHT_LIMIT:
         # Fetch more from DB if we don't have enough in the filtered set
         conn = db.get_connection()
         try:
@@ -170,7 +172,8 @@ def get_web_data(db: Database, start_ts: Optional[int] = None, end_ts: Optional[
                 params_ai.append(end_ts)
             if conditions_ai:
                 query_ai += " AND " + " AND ".join(conditions_ai)
-            query_ai += " ORDER BY start_time DESC LIMIT 15"
+            query_ai += " ORDER BY start_time DESC LIMIT ?"
+            params_ai.append(AI_HIGHLIGHT_LIMIT)
             
             cursor.execute(query_ai, params_ai)
             extra_ids = [row[0] for row in cursor.fetchall()]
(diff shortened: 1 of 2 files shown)


floze-the-genius:
Addressed the Greptile P2 in the latest commit: the regression tests now import `WEB_REPORT_SESSION_LIMIT`, derive the overflow fixture from it, and no longer hardcode `1000`/`1005`. Focused validation remains green: `9 passed`; production Ruff and `git diff --check` also pass.

bitflicker64:
> Addressed the Greptile P2 in the latest commit: the regression tests now import `WEB_REPORT_SESSION_LIMIT`, derive the overflow fixture from it, and no longer hardcode `1000`/`1005`. Focused validation remains green: `9 passed`; production Ruff and `git diff --check` also pass.

Makes sense. I’ll triage it by the end of the day. Thanks for the contribution.

bitflicker64:
@codex did i trigger you ?

termstory/web.py
```diff
@@ -242,10 +245,19 @@ def get_web_data(db: Database, start_ts: Optional[int] = None, end_ts: Optional[
     finally:
         conn.close()
 
+    total_sessions = stats["total_sessions"]
+    session_window = {
+        "limit": WEB_REPORT_SESSION_LIMIT,
+        "returned": len(sessions_data),
+        "total": total_sessions,
+        "truncated": total_sessions > len(sessions_data),
+    }
```
bitflicker64, line 254: Good metadata shape. Confirm the embedded report template actually reads `session_window` (banner like “Showing 1000 of 4500 sessions”). If not, either wire it in this PR or open a follow-up so this field isn’t unused JSON.

termstory/web.py
```diff
@@ -10,6 +10,9 @@
 
 logger = logging.getLogger(__name__)
 
+WEB_REPORT_SESSION_LIMIT = 1000
```
bitflicker64: Nice named constants. Using `LIMIT ?` with a single bound int is correct — no SQLite bind-variable explosion here (that concern applies to huge `IN (...)` lists, not this).

bitflicker64 (commented):
## Code Review Summary

**Verdict: Approve with minor suggestions** — correct fix for silent 30-session unfiltered truncation; good constants + metadata + tests.

### Intent
Unfiltered web reports were capped at 30 sessions while filtered used 1000. Unify on a named `WEB_REPORT_SESSION_LIMIT` (1000), parameterize LIMIT, and expose `session_window` metadata (issue #300).

### Impact
- Unfiltered path no longer special-cased to 30 — real bugfix.
- `AI_HIGHLIGHT_LIMIT` named constant — clarity win.
- Tests: empty window metadata, 31-session unfiltered no-truncation, and limit+5 truncated path updated to use the constant.

### What’s correct
- Single `ORDER BY ... LIMIT ?` path for both filtered and unfiltered.
- `truncated` uses stats total vs returned length; when a date range is applied, stats are overridden to the filtered counts first — so metadata stays coherent.
- Parameter binding for LIMIT avoids injection and is SQLite-safe (single bind, not a 1000-id IN list).

### Suggestions
1. **Template/UI consumption** — `session_window` is returned in JSON but (if the HTML template doesn’t read it yet) users still won’t *see* truncation. Either wire a small banner in the report UI or note a follow-up issue so metadata isn’t dead API.
2. **Optional:** also return `filtered_total` via `COUNT(*)` with the same WHERE (without LIMIT) so `total` doesn’t depend on the stats override path — more direct, same result.
3. Performance note only: loading 1000 full sessions (commands/commits) can be heavy on large DBs; acceptable for the stated product choice, but worth watching.

### Verdict
Comment / soft approve — core fix is right and well tested; template surfacing is the main follow-up.

bitflicker64:
:miku /check-star

floze-the-genius:
:miku /check-star

floze-the-genius:
Implemented the template/UI suggestion in `df859cc`: truncated reports now show a visible `Showing <returned> of <total> sessions.` status notice, while complete reports remain unchanged. The values are assigned with `textContent`, and the generated-report test covers the notice wiring. Validation: web tests `9 passed`, production Ruff and `git diff --check` green. The full local suite reached `505 passed`; its one `test_cli_error_states` foreign-key failure reproduces unchanged on a detached clean `upstream/main` worktree.

termstory/web.py
floze-the-genius: Addressed in df859cc. The embedded report now renders an aria-live status banner when session_window.truncated is true (for example, “Showing 1000 of 4500 sessions”), and tests/test_web.py asserts both the banner wiring and truncated metadata. Focused web suite passes locally.

floze-the-genius:
Rebased onto `upstream/main@1996a2f` and resolved the `termstory/web.py` conflict by retaining upstream’s unified 500-session cap through `WEB_REPORT_SESSION_LIMIT`, while preserving the PR’s parameterized query, `session_window` metadata, truncation notice, and regression coverage. Validation: `tests/test_web.py` 9 passed; `ruff check termstory/web.py` and `git diff --check` passed; full suite 524 passed with only `test_cli_error_states` failing identically on a detached clean `upstream/main@1996a2f` worktree.

bitflicker64 (approved):

bitflicker64:
Please ask the ECSoC admin to check why the bot didn’t apply the ECSoC label. If you don’t have any connections, I’ll look into it myself. Just say the word or mention me here.

PR #354, opened 2026-07-23
fix: make today comparison opt-in by default

### LINKED ISSUE #347: today --compare appears to be the default and does not render a distinct comparison view
## What I tried

I compared the normal `today` output with the documented comparison mode:

```bash
termstory today
termstory today --compare
termstory today --help
```

## What happened

`termstory today` and `termstory today --compare` rendered the same summary-style output with inline `+Xm vs yesterday` text.

The help text also says:

```text
--compare / --no-compare  Compare with yesterday's metrics [default: compare]
```

So comparison mode appears to be enabled by default, and `--compare` does not produce the side-by-side comparison described in the docs.

## Expected behavior

One of these should happen:

- `termstory today` shows a plain daily summary and `termstory today --compare` shows a clearly distinct comparison view, or
- the docs/help text should state that comparison deltas are always shown by default and users can opt out with `--no-compare`.

## Why this matters

The current behavior makes it hard to understand what the flag does, and users following the docs will expect a richer comparison than the one they get.

deepmhatre13:
Closes #347
## Summary

This PR changes the default behavior of the `today` command so that comparison with yesterday is shown only when explicitly requested using `--compare`.

## Changes

- Changed the default value of the `compare` CLI option from `True` to `False`.
- Updated the help text to clarify that comparison is optional.
- Added regression tests covering:
  - Default `today` command behavior
  - `--compare`
  - `--no-compare`

### Before

```bash
termstory today
```

Displayed today's summary along with comparison against yesterday.

### After

```bash
termstory today
```

Displays only today's summary.

```bash
termstory today --compare
```

Displays today's summary together with comparison against yesterday.

```bash
termstory today --no-compare
```

Explicitly disables comparison (same as the default behavior).

## Testing

- Added regression tests for all compare flag behaviors.
- Verified CLI help shows `--compare / --no-compare` with the default set to `no-compare`.
- ✅ `pytest tests/test_cli_commands.py -k today`
- ✅ `pytest tests/test_cli_commands.py`
- Verified only the intended files were modified.

### DIFF
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -280,7 +280,7 @@ def search_history(
 
 @app.command("today")
 def show_today(
-    compare: bool = typer.Option(True, "--compare/--no-compare", help="Compare with yesterday's metrics")
+    compare: bool = typer.Option(False, "--compare/--no-compare", help="Compare each project's time with yesterday")
 ):
     """Show today's work summary"""
     db_path = get_db_path()
--- a/tests/test_cli_commands.py
+++ b/tests/test_cli_commands.py
@@ -620,6 +620,99 @@ def test_cli_project_context_command(tmp_path, monkeypatch):
     assert "This project is for graph databases" in result_show.stdout.strip()
 
 
+def test_cli_today_compare_flag_defaults_to_false(tmp_path, monkeypatch):
+    db_file = tmp_path / "test_cli_today.db"
+    monkeypatch.setattr("termstory.cli.get_db_path", lambda: str(db_file))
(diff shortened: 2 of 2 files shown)


deepmhatre13:
@bitflicker64 I investigated the failing CI job. The failure occurs only on the Python 3.9 workflow. I ran the failing test (test_tui_generate_executive_review) locally and it passes. The changes in this PR are limited to termstory/cli.py and tests/test_cli_commands.py, and the TUI test appears unrelated. This looks like an intermittent or environment-specific failure.

deepmhatre13:
:miku /review

bitflicker64 (approved):

PR #356, opened 2026-07-24
Fix concurrent config access with atomic writes and file locking

### LINKED ISSUE #338: Concurrent first-run commands can transiently corrupt or overwrite config.json
## What I tried

I created a fresh sandbox HOME with:

- a valid `.zsh_history`
- a valid `~/.termstory/config.json` containing `"ai_enabled": true` and `"has_seen_onboarding": true`

Then I ran several ordinary read-style commands concurrently against that same HOME:

```bash
termstory stats
termstory search invoice --detailed
termstory today
termstory config get ai_enabled
```

## What happened

One command printed:

```text
Config file '.../.termstory/config.json' contains invalid data and will be ignored (Expecting value: line 1 column 1 (char 0)).
```

At the same time, `termstory config get ai_enabled` returned:

```text
False
```

The fixture config had started with `ai_enabled: true`.

## Expected behavior

Concurrent read commands should not observe a partially-written config file, and should not replace existing user settings with defaults. Config writes should be atomic and ideally guarded by a lock when first-run/default migration code may run.

## Why this matters

It is normal for users to run multiple TermStory commands from different shells, or for a TUI/report flow to overlap with a CLI command. Losing AI settings or onboarding state makes setup feel flaky and can disable features unexpectedly.

deepmhatre13:
Closes #338

## Summary

This PR fixes a race condition where concurrent TermStory processes could observe a partially-written `config.json` or overwrite existing user settings during configuration initialization or migration.

The implementation makes configuration access concurrency-safe by using atomic writes and file locking, preventing readers from seeing incomplete JSON and preserving existing configuration values under concurrent access.

## Changes

- Implemented atomic configuration writes using a temporary file followed by `os.replace()`.
- Added cross-platform file locking to synchronize concurrent configuration reads and writes.
- Ensured configuration migration remains safe and idempotent under concurrent execution.
- Added cleanup for temporary files on write failures.
- Improved error handling around configuration persistence.
- Added multiprocessing-based regression tests covering concurrent readers and writers.

## Regression Tests

Added tests to verify:

- Concurrent readers never observe invalid or partially-written JSON.
- Concurrent writers do not corrupt `config.json`.
- Existing configuration values (including `ai_enabled` and `has_seen_onboarding`) are preserved under concurrent access.
- The resulting configuration file always remains valid JSON after concurrent operations.

## Testing

- ✅ `pytest tests/test_config.py -vv`
- ✅ Added multiprocessing regression tests for concurrent configuration access.
- ✅ `ruff check .`

### DIFF
--- a/termstory/config.py
+++ b/termstory/config.py
@@ -196,13 +196,7 @@ def set_config_value(config: dict, path: str, value: Any) -> None:
     curr[parts[-1]] = value
 
 def _open_lock_file(lock_path: str) -> int:
-    """Open the lock file, retrying on transient permission errors (Windows).
-
-    Under extreme contention (many processes, short-lived locks) Windows
-    can transiently reject open() against the lock file. We retry with
-    linear backoff and jitter so callers see eventual success instead of
-    spurious permission errors.
-    """
+    """Open the lock file, retrying on transient permission errors (Windows)."""
     fd = None
     last_err = None
     for attempt in range(100):
@@ -214,7 +208,6 @@ def _open_lock_file(lock_path: str) -> int:
             if e.errno != 13:  # EACCES / Permission denied
                 raise
             import time
-            # Linear backoff with jitter, capped at 250ms
             delay = min(0.01 + (attempt * 0.005) + (random.random() * 0.01), 0.25)
             time.sleep(delay)
     raise last_err  # type: ignore[misc]
@@ -251,10 +244,6 @@ def _atomic_write_config(config_path: str, config: dict, lock_fd: int) -> None:
                     pass
     except OSError as e:
         logger.error("Failed to write config file '%s': %s", config_path, e)
-
-
-class _ConfigLockError(Exception):
-    """Raised when the config lock cannot be acquired after retries."""
 
 
 def load_config() -> dict:
@@ -318,6 +307,11 @@ def load_config() -> dict:
         _acquire_lock(fd)
     except OSError as e:
         logger.warning("Could not acquire config lock: %s", e)
+        if fd is not None:
+            try:
+                os.close(fd)
+            except OSError:
+                pass
         return dict(defaults)
 
     try:
@@ -439,14 +433,9 @@ def save_config(config: dict) -> None:
     """Save configuration dictionary to disk atomically"""
     config_path = get_config_path()
     lock_path = get_config_lock_path()
-    tmp_path = None
-    
+    fd = None
     try:
         fd = _open_lock_file(lock_path)
-    except OSError as e:
-        logger.warning("Could not acquire config lock for write: %s", e)
-        return
-    try:
         _acquire_lock(fd)
 
         dir_name = os.path.dirname(config_path)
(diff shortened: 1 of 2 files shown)


bitflicker64 (changes requested):
220-260: 🗄️ Data Integrity & Integration | 🟠 Major | 🏗️ Heavy lift

Use the same lock for reads and writes

load_config locks config_path, while save_config locks config.lock, so readers and writers never contend. On Windows, an open config.json can make os.replace() fail, and save_config currently swallows that exception, so a concurrent read can cause the write to disappear silently. Either lock the same file on both paths or drop the read lock and harden the replace step.

bitflicker64:
@kilocode /review

bitflicker64:
@kilocode /review

deepmhatre13:
@bitflicker64 kindly review the pr.

deepmhatre13:
@bitflicker64 any improvement in the pr is required?

bitflicker64:
Thanks for the solid work here, @deepmhatre13 — leaving this **open** for a maintainer decision rather than merging or closing, for a few concrete reasons:

- **Atomic writes already landed on main.** `save_config()` now writes to a temp file (`fsync` + `os.replace`), so the torn-read half of this PR is already covered. That's also the main source of the current merge conflict.
- **The remaining net-new value is the locking layer** — the cross-platform `fcntl`/`msvcrt` exclusive lock + `_deep_merge` read-modify-write, which prevents *lost updates* (two processes writing config concurrently), addressing #338 in a way atomic-write alone does not.
- **Two things block a confident merge right now:** it needs a rebase onto current main (the `save_config` rewrite conflicts), and the Windows `msvcrt.locking` retry path can't be verified in this (macOS) environment.

If you're up for rebasing onto main and trimming to just the locking + deep-merge layer on top of the existing atomic write, I think that's a clean, mergeable change. Keeping it open so that decision isn't lost.

bitflicker64:
Closing as superseded, and apologies that this sat open long enough for that to happen - the CI was fully green across all four Python versions, which is more than most PRs here manage.

The atomic-write behaviour this PR set out to add is now on `main`. `config.save_config()` writes through `tempfile.mkstemp()` in the target directory, fsyncs, then `os.replace()`s over the config path (`config.py:296-302`), which is the same core technique your branch used. #338 was closed on the back of that.

The branch has since gone to `CONFLICTING` against `main`, and rebasing it would mostly mean deleting the parts that already exist, so a fresh PR is the cleaner path if any of it is still worth landing.

One piece of your work that did *not* make it in: the file-locking layer. The merged version is atomic per write but does nothing to serialize concurrent writers, so a genuine read-modify-write race can still lose an update. If you want to revisit that specifically, it would be a welcome standalone PR.

Related and still open: I filed #402 today, which is about `save_config()` swallowing every write failure via a bare `except Exception: pass` - so the atomic write can fail completely and silently. That is adjacent to what you were already working in. Full new batch is #398 through #407.

PR #357, opened 2026-07-25
[Docs] Fix venv bootstrap and incorrect extras name in CONTRIBUTING.md

### LINKED ISSUE #335: Editable install fails in a fresh venv on macOS/Python 3.9
## What I tried

Using a fresh clone and fresh virtualenv on macOS with Python 3.9:

```bash
python3 -m venv .venv
.venv/bin/pip install -e .
```

## What happened

The editable install failed during the legacy `setup.py develop` path. The key error was:

```text
.venv/bin/python3: No module named pip
subprocess.CalledProcessError: Command '['.../.venv/bin/python3', '-m', 'pip', 'install', '-e', '.', '--use-pep517', '--no-deps']' returned non-zero exit status 1.
```

Installing non-editably worked after repairing/upgrading the venv tooling:

```bash
.venv/bin/python -m ensurepip --upgrade
.venv/bin/python -m pip install --upgrade pip setuptools wheel
.venv/bin/python -m pip install .
```

## Expected behavior

`pip install -e .` should work from a clean checkout, especially for contributors following a local development workflow.

## Environment

- macOS
- Python 3.9
- Fresh clone of `bitflicker64/Termstory`
- TermStory package version observed: `0.6.5` in `pyproject.toml`

## Why this matters

This blocks the first contributor/developer setup path before anyone can run tests or try changes locally.

dharapandya85:
## Summary
Fixes contributor setup instructions in CONTRIBUTING.md by adding virtual environment creation and activation steps, and updates the setuptools minimum version requirement in pyproject.toml. This addresses issue #335 where the original instructions referenced a non-existent `.[dev]` extra and skipped venv creation, causing failures for contributors with outdated pip bootstraps.

## Changes
- **Documentation**: CONTRIBUTING.md now includes venv creation (`python3 -m venv .venv --upgrade-deps`) and activation steps before the editable install, and changes the dependency extra from `.[dev]` to `.[test]`
- **Build System**: pyproject.toml raises the minimum setuptools requirement from `>=61.0.0` to `>=64`

## Fixes
- Fixes #335 — CONTRIBUTING.md now creates a venv before installation and uses the existing `.[test]` extra instead of the non-existent `.[dev]`

## Testing
After applying these changes, verify the setup works by running:
```bash
python3 -m venv .venv --upgrade-deps
source .venv/bin/activate
pip install -e ".[test]"
python3 -m pytest tests/ -v
```

## Notes
The diff leaves the original `pip install -e ".[dev]"` line intact before the venv creation step. Review feedback indicates this stale line should be removed since `.[dev]` does not exist in pyproject.toml's optional dependencies (only `test` and `rag` are defined). Contributors should follow the new venv creation and activation steps, then use `.[test]` for installation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,8 +13,11 @@ Please read our [Code of Conduct](CODE_OF_CONDUCT.md) before contributing to hel
    cd termstory
    ```
 2. Install the package in editable mode with development dependencies:
-   ```
-
+   ```bash
+   python3 -m venv .venv --upgrade-deps
+   source .venv/bin/activate       # on Windows: .venv\Scripts\activate
+   pip install -e ".[test]"
+    ```
 ## Running Tests
 
 To run the test suite, simply use pytest:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,9 @@ Please read our [Code of Conduct](CODE_OF_CONDUCT.md) before contributing to hel
 2. Install the package in editable mode with development dependencies:
    ```bash
    pip install -e ".[dev]"
+   python3 -m venv .venv --upgrade-deps
+   source .venv/bin/activate       # on Windows: .venv\Scripts\activate
+   pip install -e ".[test]"
    ```
 
 ## Running Tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,11 +13,6 @@ Please read our [Code of Conduct](CODE_OF_CONDUCT.md) before contributing to hel
    cd termstory
    ```
 2. Install the package in editable mode with development dependencies:
-   ```bash
-   pip install -e ".[dev]"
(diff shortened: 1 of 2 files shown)


dharapandya85:
@bitflicker64, please review.

bitflicker64 (changes requested):
@dharapandya85, please take care of the reviewers' suggested changes. If any look incorrect, just comment on why, okay?

bitflicker64:
@Kilo review

bitflicker64:
@greptile review 
@coderabbit review

bitflicker64 (changes requested):
**Bootstrap still runs too late**

When a contributor has missing or broken system pip, the preceding `pip install -e ".[dev]"` runs before `.venv` is created and activated, causing setup to fail before the new bootstrap commands are reached. That preceding command also still names the undefined `dev` extra, so it should be replaced by this venv-first sequence rather than retained ahead of it.

PR #370, opened 2026-07-28
fix: validate config values in config set

### LINKED ISSUE #342: config set accepts invalid types and silently coerces unknown booleans
## What I tried

I used the public config CLI to set a numeric field and a boolean field to invalid values:

```bash
termstory config set request_timeout_seconds abc
termstory config get request_timeout_seconds

termstory config set ai_enabled maybe
termstory config get ai_enabled
```

## What happened

The timeout accepted a non-numeric value:

```text
Set config key 'request_timeout_seconds' to 'abc'
abc
```

The boolean value `maybe` was silently coerced to `False`:

```text
Set config key 'ai_enabled' to 'False'
False
```

After further setup attempts, config could show a confusing state like:

```text
ai_enabled: True
active_provider: disabled
request_timeout_seconds: abc
```

and `termstory ask ...` still reported:

```text
AI capabilities are currently disabled.
```

## Expected behavior

`config set` should validate known config keys according to their expected type/range. For example:

- `request_timeout_seconds` should require a positive number
- boolean keys should accept only clear boolean inputs such as `true/false`, `1/0`, `yes/no`
- setting `ai_enabled=true` while `active_provider=disabled` should either warn or guide the user to choose a provider

## Why this matters

Configuration is the main path to enabling AI. Silent coercion and invalid values make it easy to end up with a config file that looks enabled but behaves disabled.

deepmhatre13:
Closes #342

## Summary

This PR adds strict validation to the `termstory config set` command to prevent invalid configuration values from being written while preserving existing behavior for valid inputs.

Previously, invalid numeric values (e.g. `abc` for `request_timeout_seconds`) were stored directly, and unknown boolean values (e.g. `maybe`) were silently coerced to `False`, leading to confusing configuration states.

### Validation

- Validate numeric configuration keys (e.g. `request_timeout_seconds`)
  - Accept positive integers and floating-point values
  - Reject non-numeric, zero, negative, `NaN`, and infinite values
  - Display clear error messages without modifying the configuration

- Validate boolean configuration keys
  - Accept only:
    - `true`
    - `false`
    - `yes`
    - `no`
    - `1`
    - `0`
  - Case-insensitive
  - Reject ambiguous values such as `maybe`, `on`, `off`, `enable`, and `disable`

### Configuration behavior

- Preserve API keys, tokens, and passwords as raw strings
- Preserve unknown/custom configuration keys without validation
- Apply validation after legacy key translation so legacy configuration names continue to work correctly

### User guidance

- When enabling `ai_enabled=true` while `active_provider=disabled`, display a helpful warning explaining that a provider still needs to be configured instead of silently leaving the user in a confusing state.

## Tests

Added tests covering:

- Valid numeric values
- Invalid numeric values
- Valid boolean values
- Invalid boolean values
- Legacy key validation
- API key leading-zero preservation
- Unknown key preservation
- AI enablement guidance
- Configuration immutability after validation failures

## Verification

- ✅ `ruff check .`
- ✅ `pytest tests/test_cli_commands.py -v`
- ✅ `pytest tests/test_config.py -v`

### DIFF
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -1377,41 +1377,84 @@ def config_set(key: str, value: str):
     from termstory.config import load_config, save_config, set_config_value, get_config_value
     config = load_config()
     
-    current_val = get_config_value(config, key)
-    if isinstance(current_val, bool):
-        converted_value = value.lower() in ("true", "1", "yes")
-    elif isinstance(current_val, int):
-        try:
-            converted_value = int(value)
-        except ValueError:
-            converted_value = value
-    elif isinstance(current_val, float):
-        try:
-            converted_value = float(value)
-        except ValueError:
+    from termstory.config import translate_legacy_key
+    effective_key = translate_legacy_key(config, key)
+
+    KNOWN_DEFAULTS = {
+        "ai_enabled": False,
+        "active_provider": "disabled",
+        "request_timeout_seconds": 30,
+        "ai_max_failures": 3,
+        "ai_cooldown_seconds": 60.0,
+        "has_seen_onboarding": False,
+        "has_seen_timestamp_prompt": False,
+        "has_seen_onboarding_reminder": False,
+        "max_history_age": 5,
+        "max_query_log": 10000,
+        "db_timeout": 30.0,
+        "reminder_poll_interval": 300,
+        "clustering_threshold": 0.6,
+        "nfs_timeout_cache_ttl": 60,
+    }
+
+    default_val = KNOWN_DEFAULTS.get(effective_key)
+    if default_val is not None:
+        if isinstance(default_val, bool):
+            lowered = value.lower()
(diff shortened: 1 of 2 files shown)


deepmhatre13:
gitar auto-apply:on

deepmhatre13:
@bitflicker64 I investigated the remaining CI failure.

My changes in this PR are limited to `termstory/cli.py` and `tests/test_cli_commands.py`. The CLI validation tests pass locally, and the PR now passes on Python 3.9, 3.11, and 3.12.

The remaining failure is in `tests/test_tui.py::test_tui_bulk_auto_summarize`, which exercises the TUI summarization flow and appears unrelated to the CLI config validation changes introduced in this PR. I wasn't able to reproduce a connection between this failure and the changes in this branch.

bitflicker64 (approved):

PR #374, opened 2026-07-30
fix(tui): use configured api_base_url default in AI settings modal

### LINKED ISSUE #152: [Medium] cli.py: Hardcoded Groq API base URL should be configurable
## Location

| Field | Value |
|-------|-------|
| **File** | `termstory/cli.py` |
| **Line** | 699 |
| **Type** | `FIXME` |
| **Category** | `design` |
| **Difficulty** | Medium |

---

## Analysis

Let me read the relevant file to analyze the FIXME comment at line 699.

Let me look at more context around this area, including the config loading and any default URL constants defined elsewhere.

Let me read the config.py and ai.py files to understand the full context, and also look at more of cli.py around line 699.

Now I have a comprehensive understanding of the issue. Let me analyze what's happening:

1. In `cli.py` around line 699, there's a block that hardcodes fallback API base URLs for each provider (groq, openai, ollama) when `api_base_url` is empty.

2. However, `config.py` already has these exact same defaults baked into the `load_config()` function's `defaults` dict (lines 144-159), and `load_config()` also calls `merge_defaults()` which recursively fills in missing keys from defaults.

3. The problem: the `config.get()` call on line 695 uses a dot-path string like `"providers.groq.api_base_url"`, but `config` is a plain Python dict returned by `load_config()`. The `dict.get()` method does NOT support dot-notation path traversal — it would look for a literal key `"providers.groq.api_base_url"` in the top-level dict, which doesn't exist. So `api_base_url` will always be empty string `""`, and the fallback block at lines 698-704 will always trigger.

Wait, let me re-read. The config object returned by `load_config()` is a plain dict. `config.get("providers.groq.api_base_url", ...)` would look for a literal key `"providers.groq.api_base_url"` in the top-level dict, which doesn't exist. So it falls back to `config.get("groq_api_base_url", "")` which also doesn't exist (legacy flat key). So `api_base_url` will always be `""`.

But wait — `load_config()` already has defaults that include `providers.groq.api_base_url = "https://api.groq.com/openai/v1"`. The proper way to access this would be `config["providers"]["groq"]["api_base_url"]` or using `get_config_value(config, "providers.groq.api_base_url")`.

So the root cause is:
1. `config.get()` doesn't support dot-notation, so the config lookup always fails and returns `""`
2. The fallback hardcoded URLs in `cli.py` lines 698-704 are a workaround for this broken lookup
3. The proper fix is to use `get_config_value(config, f"providers.{provider}.api_base_url")` from `config.py`, which already handles dot-path traversal AND legacy key translation
4. Since `load_config()` already merges defaults (including the correct URLs), using `get_config_value()` would return the right URL from config without any hardcoding needed

The FIXME comment mentioned in the question is described as being at line 699, but there's no actual `# FIXME` comment in the code — the FIXME is the conceptual issue the user is asking about. The hardcoded fallback block at lines 698-704 is the problem.

Let me also check if there are other places in cli.py that do the same pattern (config.get with dot-notation).
Based on the full source read, here is the analysis:

---

## === CONTEXT ===

`termstory/cli.py`, lines 692–705:

```python
if ai_enabled and provider != "disabled":
    from termstory.ai import translate_git_anger
    api_key = config.get(f"providers.{provider}.api_key", config.get(f"{provider}_api_key", ""))
    api_base_url = config.get(f"providers.{provider}.api_base_url", config.get(f"{provider}_api_base_url", ""))
    model_name = config.get(f"providers.{provider}.model_name", config.get(f"{provider}_model_name", ""))

    if not api_base_url:                                    # line 698
        if provider == "groq":                              # line 699
            api_base_url = "https://api.groq.com/openai/v1"
        elif provider == "openai":
            api_base_url = "https://api.openai.com/v1"
        elif provider == "ollama":
            api_base_url = "http://localhost:11434/v1"
```

---

## === ROOT CAUSE ===

There are two compounding problems. First, `config` is a plain Python `dict` returned by `load_config()`, and `dict.get()` does **not** support dot-notation path traversal. The call `config.get(f"providers.{provider}.api_base_url", ...)` looks for a literal top-level key `"providers.groq.api_base_url"`, which never exists, so it always falls back to `""`. The correct accessor is `get_config_value(config, path)` from `termstory/config.py`, which already handles dot-path traversal and legacy key translation.

Second, the hardcoded fallback block at lines 698–704 is entirely redundant: `load_config()` already defines these exact same URLs as defaults and merges them into the config dict via `merge_defaults()` before returning.

So the real bug is that the dot-path config lookup is silently broken everywhere in `cli.py` that uses `config.get(f"providers.{provider}.*")`, causing the code to always fall through to the hardcoded strings instead of reading the user's configured (or default) values. This means a user who changes their Groq base URL in `config.json` will have their setting silently ignored in this code path.

---

## === SUGGESTED FIX ===

- **FILE: `termstory/cli.py`, LINE 694** — Replace the three `config.get(f"providers.{provider}.*")` calls with `get_config_value(config, f"providers.{provider}.api_key")` etc., importing `get_config_value` from `termstory.config`. This makes the lookup actually traverse the nested dict.

- **FILE: `termstory/cli.py`, LINES 698–704** — Delete the entire `if not api_base_url:` fallback block. It is unnecessary once the lookup is fixed, because `load_config()` already merges the correct default URLs into the config before returning.

- **FILE: `termstory/cli.py`, LINE 690** — The provider is read as `config.get("ai_provider", "disabled")`, but the config key was migrated to `"active_provider"`. This should be `config.get("active_provider", "disabled")` (or use `get_config_value`), otherwise the provider will always resolve to `"disabled"` on non-legacy configs.

- **FILE: `termstory/cli.py`** — Audit all other occurrences of `config.get(f"providers.{provider}.*")` (there are 9 total matches in the file) and apply the same fix consistently.

---

## === EFFORT ===

**S** — The fix is mechanical: swap `config.get(f"providers.{provider}.X", ...)` for `get_config_value(config, f"providers.{provider}.X")` at each call site, delete the now-dead fallback block, and fix the `ai_provider` → `active_provider` key name. No architectural changes needed.

---

## === PREREQUISITES ===

- Understand how `load_config()` / `get_config_value()` / `set_config_value()` work in `termstory/config.py` — specifically that config is a nested dict and dot-paths must be traversed explicitly.

- Know the legacy key migration logic (`ai_provider` → `active_provider`, `groq_api_key` → `providers.groq.api_key`) so you don't accidentally break backward compatibility when fixing the key names.

muditagrawal2007:
The TUI's "Save" action in the AI provider settings modal used a hardcoded fallback block (mirroring the one previously removed from cli.py:699) for groq/openai/ollama base URLs when the user left the "Base URL" field blank. That ignored any custom value set via `termstory config set providers.<provider>.api_base_url <url>` and silently overwrote the user's setting on save.

Replace the hardcoded fallback with
`get_config_value(config, f"providers.{provider}.api_base_url")`, which already returns the configured default (merged into the nested config by load_config()) — so a user-supplied URL is preserved and a blank field correctly resolves to the provider's configured default.

Adds a regression test in tests/test_config.py that pins a custom api_base_url in config.json and asserts it survives both `get_config_value` and `cli.get_ai_provider_settings`, preventing future regressions of the dot-notation lookup bug.

Fixes #152

## Description
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

## Type of Change
Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist:
- [ ] My code follows the "density over decoration" philosophy.
- [ ] I have not used `rich.panel.Panel` in my code.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

### DIFF
--- a/termstory/tui.py
+++ b/termstory/tui.py
@@ -585,7 +585,7 @@ class OnboardingScreen(_DeferredDismissMixin, ModalScreen[dict]):
             base_url = self.query_one("#input-base-url").value.strip()
             model_name = self.query_one("#input-model-name").value.strip()
             github_username = self.query_one("#input-github-username").value.strip().lstrip('@')
-            
+
             error_label = self.query_one("#error-api-key")
             if self.selected_provider in ("groq", "openai", "custom") and not api_key:
                 error_label.update("API Key cannot be empty.")
@@ -593,15 +593,19 @@ class OnboardingScreen(_DeferredDismissMixin, ModalScreen[dict]):
                 return
             else:
                 error_label.styles.display = "none"
-            
+
             if not base_url:
-                if self.selected_provider == "groq":
-                    base_url = "https://api.groq.com/openai/v1"
-                elif self.selected_provider == "openai":
-                    base_url = "https://api.openai.com/v1"
-                elif self.selected_provider == "ollama":
-                    base_url = "http://localhost:11434/v1"
-                    
+                # Fall back to the configured default for this provider instead
+                # of a hardcoded URL. ``load_config()`` already merges the
+                # default endpoints into the nested ``providers.*.api_base_url``
+                # keys, so this picks up customisations set via
+                # ``termstory config set providers.groq.api_base_url <url>``.
+                # Fixes #152.
+                from termstory.config import get_config_value
+                base_url = get_config_value(
+                    self.config, f"providers.{self.selected_provider}.api_base_url"
+                ) or ""
+
             if "providers" not in self.config:
                 self.config["providers"] = {}
             if self.selected_provider not in self.config["providers"]:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,6 +5,54 @@ from unittest.mock import patch, mock_open
 
 from termstory.config import load_config, save_config, get_config_path
 
+def test_api_base_url_is_configurable_per_provider(tmp_path, monkeypatch):
+    """#152: the api_base_url for every provider must come from config and
+    be overridable via ``termstory config set providers.<p>.api_base_url``.
+    Hardcoded fallbacks (the old behaviour in cli.py / tui.py) silently
+    ignored user customisations.
+    """
+    config_file = tmp_path / "config.json"
+    config_data = {
+        "active_provider": "groq",
+        "providers": {
+            "groq": {
+                "api_key": "fake-key",
+                # User overrides Groq to point at a self-hosted proxy
+                "api_base_url": "https://my-proxy.example.com/openai/v1",
+                "model_name": "llama-3.1-8b-instant",
+            },
+        },
+    }
+    config_file.write_text(json.dumps(config_data))
+
+    with patch("termstory.config.get_config_path", return_value=str(config_file)):
+        config = load_config()
+
+        # get_config_value must use dot-path traversal (dict.get() does NOT)
+        from termstory.config import get_config_value
+
+        url = get_config_value(config, "providers.groq.api_base_url")
+        assert url == "https://my-proxy.example.com/openai/v1", (
+            "Configured api_base_url was not returned by get_config_value() "
+            "(regression of #152 dot-notation lookup bug)"
+        )
+
+        # Other providers must still return their defaults after merge
+        openai_url = get_config_value(config, "providers.openai.api_base_url")
+        assert openai_url == "https://api.openai.com/v1"
+
+        ollama_url = get_config_value(config, "providers.ollama.api_base_url")
+        assert ollama_url == "http://localhost:11434/v1"
+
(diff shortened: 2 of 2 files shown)


bitflicker64 (approved):
LGTM — makes the AI settings modal respect the configured api_base_url default.

PR #386, opened 2026-07-30
fix(ask): use COALESCE for end_time in session matching

### LINKED ISSUE #298: [Medium] ask.py: session matching uses s.end_time + 600 without COALESCE — active sessions missed
## Bug Description

Session retrieval for `termstory ask` joins commits / FTS hits with:

```sql
... <= s.end_time + 600
```

Active (open) sessions can have `end_time IS NULL` in SQLite. In SQL, `NULL + 600` is NULL, so the comparison fails and those sessions are **excluded** from ask context even when their commands match the query.

`termstory/search.py` and `termstory/database.py` already use the safe form:

```sql
COALESCE(s.end_time, s.start_time) + 600
```

`ask.py` does not — retrieval quality diverges from search.

## Location

`termstory/ask.py`:

1. FTS path (~line 159):
   ```sql
   AND CAST(f.timestamp AS INTEGER) <= s.end_time + 600)
   ```
2. LIKE fallback path (~line 199):
   ```sql
   AND co.timestamp <= s.end_time + 600
   ```

## Steps to Reproduce

1. Ensure an open/active session exists with `end_time` NULL and at least one distinctive command.
2. Run `termstory ask "<word unique to that command>"` (or call the retrieval helper used by ask).
3. Actual: session often missing from context when the only match is via commit window / FTS commit linkage.
4. Expected: same COALESCE semantics as `search.py`.

## Expected Behavior

Use `COALESCE(s.end_time, s.start_time) + 600` (and the matching start-side buffer already present) in both SQL paths.

## Suggested Fix

Align both SQL fragments with `search.py` lines ~171 and ~224. Add a unit test that inserts a session with `end_time=NULL`, a matching command/commit, and asserts the session id is returned by the ask retrieval helper.

## Acceptance Criteria

- [ ] Both ask SQL paths use COALESCE (or equivalent Python-side effective end)
- [ ] Active session with NULL `end_time` is still discoverable
- [ ] Regression test in `tests/test_ask.py`

muditagrawal2007:
Use COALESCE(s.end_time, s.start_time) + 600 in both SQL paths in termstory/ask.py so active sessions (end_time IS NULL) are not excluded from ask context.

Closes #298

### DIFF
--- a/termstory/ask.py
+++ b/termstory/ask.py
@@ -196,7 +196,7 @@ def search_ask(query: str, db) -> List[Session]:
                 LEFT JOIN commands c ON s.id = c.session_id
                 LEFT JOIN commits co ON s.project_id = co.project_id
                     AND co.timestamp >= s.start_time - 300
-                    AND co.timestamp <= s.end_time + 600
+                    AND co.timestamp <= COALESCE(s.end_time, s.start_time) + 600
                 WHERE {" OR ".join(where_clauses)}
             """
             cursor.execute(sql, params)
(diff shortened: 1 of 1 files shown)


bitflicker64 (approved):
LGTM — COALESCE fix for null end_time is correct and safe.

PR #388, opened 2026-07-30
feat: make macOS install.log path configurable

### LINKED ISSUE #150: [Medium] timestamp_detective.py: Hardcoded macOS log path /private/var/log/install.log s…
## Location

| Field | Value |
|-------|-------|
| **Line** | 1066 |
| **Type** | `FIXME` |
| **Category** | `design` |
| **Difficulty** | Medium |

---

## Analysis

Let me read the relevant file and search for the FIXME comment to understand the context.

Let me read more context around the function, including how it's called and whether there's any configuration mechanism in the codebase.

Let me look at how `detect_macos_syslog` is called within the file, and also check the `__init__` method and config.py for any existing configuration patterns that could be used.

Let me look at the `detect_macos_system_log` method (which calls `detect_macos_syslog`) and the `__init__` method to understand the full picture, and also check the test file for the relevant tests.

I have enough context to provide a thorough analysis. Let me synthesize what I found.

Key findings:
1. Line 1066: `log_path = "/private/var/log/install.log"` is hardcoded inside `detect_macos_syslog()`
2. `TimestampDetective.__init__` accepts `search_root` and `project_paths` but no `macos_log_path` parameter
3. The config system in `config.py` has a `defaults` dict with various settings, but no entry for the macOS log path
4. The `detect_macos_syslog` method is called by `detect_macos_system_log`, which is registered in `_run_all_detectors`
5. The `TimestampDetective` is instantiated in `parser.py` at line 253-256 with only `search_root` and `project_paths`

The fix involves:
1. Adding `macos_install_log_path` as a constructor parameter to `TimestampDetective.__init__` with a default of `"/private/var/log/install.log"`
2. Storing it as `self.macos_install_log_path`
3. Using `self.macos_install_log_path` instead of the hardcoded string in `detect_macos_syslog`
4. Optionally adding it to `config.py` defaults and threading it through from `parser.py`

The effort is S (a few hours) since it's a straightforward parameter extraction.

=== CONTEXT ===

```python
# termstory/timestamp_detective.py, lines 1055–1073
def detect_macos_syslog(self, package: str) -> Optional[int]:
    """
    Helper — scan macOS `/private/var/log/install.log` for an install record
    of `package` (or any brew-related line) and return the most recent
    matching install timestamp.

    Lines look like:
        2024-01-01 12:00:00+00 host installd[1]: ... Installed: <formula>

    Returns the Unix timestamp or None.
    """
    log_path = "/private/var/log/install.log"   # ← line 1066, hardcoded
    if not os.path.exists(log_path):
        return None
    try:
        with open(log_path, "r", encoding="utf-8", errors="ignore") as f:
            lines = f.readlines()
    except Exception:
        return None
```

---

=== ROOT CAUSE ===

The path `"/private/var/log/install.log"` is a string literal baked directly into `detect_macos_syslog`. The `TimestampDetective.__init__` already accepts two configurable parameters (`search_root`, `project_paths`) stored as instance attributes, but no equivalent parameter exists for the macOS log path.

The project-wide config system in `config.py` has a `defaults` dict that covers AI providers, history age, and query limits, but contains no entry for this path either.

As a result, users on non-standard macOS setups (e.g., a relocated log via `asl`, a Docker-based macOS environment, or a test harness) cannot override the path without patching source code, and the test suite must monkey-patch `os.path.exists` rather than simply passing an alternate path to the constructor.

---

=== SUGGESTED FIX ===

- **FILE:** `termstory/timestamp_detective.py`, **LINE: 92** — Add `macos_install_log: str = "/private/var/log/install.log"` to `__init__`'s signature and store it as `self.macos_install_log = macos_install_log`.

- **FILE:** `termstory/timestamp_detective.py`, **LINE: 1066** — Replace the literal `log_path = "/private/var/log/install.log"` with `log_path = self.macos_install_log`.

- **FILE:** `termstory/config.py`, **LINE: 165** (inside `defaults`) — Add `"macos_install_log": "/private/var/log/install.log"` so the path can be persisted and overridden via the user config file.

- **FILE:** `termstory/parser.py`, **LINE: 253** — When constructing `TimestampDetective`, read the config value and pass it as `macos_install_log=config.get("macos_install_log", "/private/var/log/install.log")`.

- **FILE:** `tests/test_timestamp_detective.py` — Update the `detect_macos_syslog` tests to pass a temp-file path via the constructor instead of patching `os.path.exists`, which will make the tests simpler and validate the new parameter end-to-end.

---

=== EFFORT ===

**S** (a few hours)

---

=== PREREQUISITES ===

- Familiarity with Python dataclasses/`__init__` patterns and how `config.py`'s `load_config` / `get_config_value` helpers work in this project.

- Understanding that `/private/var/log/install.log` is the macOS `installd` log (written by the system installer daemon), distinct from the Homebrew-specific logs under `$(brew --prefix)/var/log/`. No special macOS entitlements are needed to read it on a standard developer machine.

muditagrawal2007:
Add macos_install_log constructor parameter to TimestampDetective, default it in config.py, and thread through parse_zsh_history() so users can override the path.

Closes #150

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -103,9 +103,7 @@ Toggle observability settings (such as Nemo relay / DeepWiki settings) for Herme
 ```bash
 termstory                    # Today's timeline
 termstory today              # Same as above
-termstory today --detailed   # All commands, no noise filtering, with timestamps
 termstory today --compare    # Side-by-side with yesterday
-termstory today --stats      # Command category frequency table
 ```
 
 ### Search
@@ -119,26 +117,10 @@ termstory search docker --detailed
 termstory search docker --semantic   # Local hybrid semantic/RAG search
 ```
 
-### Historical
-
-```bash
-termstory week               # Current week
-termstory week --last        # Previous week
-termstory month              # Current month
-termstory month "May 2026"   # Specific month
-termstory month --last       # Previous month
-```
-
 ### Projects
 
 ```bash
 termstory project <name>                       # 30-day deep dive
-termstory project myapp --files                # Files edited, by frequency
-termstory project myapp --stats                # Command category breakdown
-termstory projects                             # All tracked projects
-termstory projects --sort time                 # By total hours (default)
-termstory projects --sort recent               # By last active date
-termstory projects --sort name                 # Alphabetically
 termstory project context <name> "description" # Set goals/context for a project
 termstory project context <name> --show        # View project context description
 ```
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -10,7 +10,7 @@ termstory ui   # Fresh start, re-runs onboarding
 ### Force a re-ingest
 
 ```bash
-termstory today --detailed
+termstory today
 ```
 
 Reads the history file fresh and updates the database.
--- a/termstory/agy.py
+++ b/termstory/agy.py
@@ -1,0 +1,378 @@
+"""
+agy.py — AI Pair Programmer Bridge
+
+Launches ``agy -p`` (the AI pair-programmer CLI) from within TermStory,
+bridging the developer's recent shell-history context into the live AI
+session so the assistant starts warm instead of cold.
+
+Design
+------
+* **Context gathering** — pulls the most recent N commands, the current
+  project name/path, and recent git commit messages from the TermStory
+  SQLite database.
+* **Privacy-first** — every command and commit message is run through
+  :func:`termstory.sanitizer.redact_command` before it leaves the local
+  machine.  Sessions containing blacklisted commands (vault logins,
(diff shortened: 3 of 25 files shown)


bitflicker64:
@muditagrawal2007 lol pr fuck up hogyi sorry 😭

bitflicker64:
Closing this one. The PR title says "make macOS install.log path configurable," but the diff spans 25 files (+2418/-214) and introduces unrelated new subsystems (`termstory/agy.py`, `test_agy.py`, `test_tui_cyberpunk_animations.py`, large `tui.py`/`project.py`/`search.py` changes). A change set this broad and unrelated to its stated scope can't be reviewed or merged as a single unit. Happy to review focused PRs — one self-contained change each — if you'd like to resubmit.

PR #395, opened 2026-08-01
feat(anger-translator): extend lookback to 3h and detect kill -9 / rapid-recompile signals (#37)

### LINKED ISSUE #37: The 'Git-Blame' Anger Translator 🤬
Instead of just showing your clean, generic "fix stuff" commit messages, the AI cross-references them with the preceding 3 hours of frantic shell errors, furious `kill -9` commands, and rapid recompiles. It then generates the *real*, dramatically unfiltered commit message describing what you were actually feeling (e.g., "fix typo" translates to "I spent 3 hours debugging because I forgot a single trailing slash in a config file and my soul has left my body.").

muditagrawal2007:
Closes #37

## Summary

The existing `anger-translator` implementation already cross-references git commits with their preceding terminal errors, but it has two gaps vs the issue's spec:

1. **Lookback window** is hardcoded to 30 minutes (`ts - 1800`); the issue asks for "the preceding 3 hours of frantic shell errors".
2. **No explicit signal detection** for the two specific things the issue calls out — "furious `kill -9` commands" and "rapid recompiles". Today the code just pulls all `exit_code != 0` commands without flagging these patterns.

This PR closes both gaps.

## Changes

**`termstory/cli.py`**
- New `--lookback-hours` option (default `3.0`, matching the issue spec).
- New helper `_is_recompile_command()` — whole-word matcher for ~25 build/compile invocations (make, cargo build, go build, gcc, g++, tsc, ninja, webpack, mvn, gradle, dotnet, swift, xcodebuild, bazel, etc.). Defensive against false positives like `makeup` or `cmake-format`.
- New helper `_count_rapid_recompile_clusters()` — counts how many "rapid-recompile" clusters appear in a list of commands, defined as `>= 3` build/compile invocations within any 60-second sliding window. 60s is the inclusive boundary; 61s is not.
- `anger_translator` now counts `kill -9` / `kill -SIGKILL` invocations and rapid-recompile clusters in the lookback window and threads them into each commit's data dict as `kill_count` and `recompile_clusters` fields.

**`termstory/ai.py`**
- `translate_git_anger` includes a "Detected Aggression Signals" block in the LLM prompt when either signal fires, so the model can roast on the specific signals the issue calls out rather than just generic preceding errors.

**`termstory/formatter.py`**
- `format_anger_translation_heuristics` gets three new branches (in order of specificity):
  - `kill_count >= 2` → 💀 FURY / PROCESS WHACK-A-MOLE
  - `recompile_clusters >= 1` and `kill_count == 0` → 🔥 RECOMPILING IN A PANIC
  - `kill_count >= 1` and `recompile_clusters >= 1` → 🌋 BUILD BROKE & PROCESSES DIED
- The existing error-count branches (RAGE & DESPAIR / FRUSTRATION / TRIUMPH) are preserved as the fallback when neither signal fires.
- Adds a small signal badge (`[kill -9 x3, recompile-clusters x1, fails x5]`) so the user can see *why* the chosen roast was chosen.

**`tests/test_git_blame_anger_fortune_teller.py`**
- 15 new tests covering: `_is_recompile_command` positive + no-false-positives, `_count_rapid_recompile_clusters` for single/two/zero/mixed-builders/60s-boundary/mismatched-lengths, LLM prompt contents (kill -9 + recompile), and the three new heuristic branches plus the error-count fallback and backwards-compat for callers that don't pass the new keys.

## Test plan

- [x] New tests pass locally (31/31 in `test_git_blame_anger_fortune_teller.py`)
- [x] Full project test suite: 665 passed, 1 pre-existing failure in `test_cli_restore_confirmation.py::test_restore_dry_run_does_not_modify_live_db` (Rich line-wrapping on long temp paths, also fails on `upstream/main` — unrelated)
- [x] Baseline comparison via worktree: my branch adds 17 passing tests vs `upstream/main`, with 0 regressions
- [x] `ruff check` shows 0 new issues in the lines added by this PR (baseline = 231, mine = 231)

## Backwards compatibility

- The new `kill_count` and `recompile_clusters` fields default to 0 when not supplied (`item.get("kill_count", 0)`), so any existing caller that only passes `preceding_errors` continues to work unchanged.
- The error-count branches are unchanged in behavior — they only fire when the new aggression-signal branches do not.

## Out of scope

The `fix`/`bug`/`crash`/`issue` commit-message override (which rewrites the emotion to 🩹 EXHAUSTION / PATCHING SHIT) is preserved as-is. It sits at the bottom of the chain and always wins, so a commit message containing "fix" + heavy kill signals will still show as EXHAUSTION. The test suite uses non-fix commit messages to exercise the new branches cleanly.

### DIFF
--- a/termstory/ai.py
+++ b/termstory/ai.py
@@ -882,9 +882,23 @@ def translate_git_anger(
         commit_msg = item.get("message", "")
         errors = item.get("preceding_errors", [])
         error_lines = "\n".join(f"  - [FAIL] {err}" for err in errors) if errors else "  - (No preceding terminal errors detected)"
+
+        # Issue #37 calls out specific signals: furious kill -9 commands and rapid
+        # recompiles. Surface them in the prompt so the LLM can roast on them.
+        signal_lines = []
+        if item.get("kill_count", 0) > 0:
+            signal_lines.append(f"  - {item['kill_count']}x 'kill -9' / SIGKILL command(s) detected")
+        if item.get("recompile_clusters", 0) > 0:
+            signal_lines.append(
+                f"  - {item['recompile_clusters']} rapid-recompile cluster(s) "
+                f"(>=3 build/compile invocations within 60s)"
+            )
+        signal_block = ("Detected Aggression Signals:\n" + "\n".join(signal_lines)) if signal_lines else ""
+
         formatted_blocks.append(
             f"Commit: {commit_msg} ({commit_hash})\n"
             f"Preceding Errors:\n{error_lines}"
+            + (f"\n{signal_block}" if signal_block else "")
         )
     
     payload = "\n\n".join(formatted_blocks)
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -645,25 +645,87 @@ def predict_cmd(
     console.print(Text.from_ansi(output))
 
 
+# Build/compile invocations that count toward a "rapid recompile" signal.
+# Matched as whole-word tokens against the command line (case-insensitive).
+_RECOMPILE_PATTERNS = (
+    "make", "cmake", "ninja", "cargo build", "rustc", "go build",
+    "gcc", "g++", "clang", "clang++", "tsc", "ts-node", "webpack",
+    "vite build", "rollup", "esbuild", "swift build", "xcodebuild",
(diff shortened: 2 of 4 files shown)


bitflicker64:
@greptile review

muditagrawal2007 (commented):
done

PR #408, opened 2026-08-07
[Date] Honor TERMSTORY_DATE_OVERRIDE at remaining datetime.now() site…

### LINKED ISSUE #407: TERMSTORY_DATE_OVERRIDE is honoured inconsistently: 17 datetime.now() call sites bypass get_current_time()
## Summary

`date_utils.get_current_time()` exists so that `TERMSTORY_DATE_OVERRIDE` can pin "now" for tests and for reproducing time-dependent behaviour. Most of the codebase routes through it. Seventeen call sites do not, and call `datetime.now()` directly, so the override applies to some parts of a single command's output and not others.

## Evidence

```
$ grep -rn "datetime\.now()" termstory/*.py | grep -v date_utils.py
termstory/backup.py:30
termstory/database.py:735
termstory/parser.py:248,260,446,477,738,771,821
termstory/predict.py:313
termstory/timeline.py:57
termstory/web.py:210
termstory/tui.py:29
termstory/timestamp_detective.py:112,1087
termstory/exporter.py:22,33
```

Not all of these are wrong. `exporter.py:22` and `:33` compute a UTC offset for a Windows fallback and genuinely want wall-clock time. `tui.py:29` is a debug log line. The rest look like they should be honouring the override:

- `timestamp_detective.py:112` sets `self.now`, the anchor the whole forensic recovery is measured against
- `parser.py` uses it as the ceiling for synthetic legacy timestamps
- `predict.py:313`, `timeline.py:57`, `web.py:210` set the reference point for their entire output
- `database.py:735` writes a timestamp into a row

## Why this matters

Two concrete consequences. First, a test that sets `TERMSTORY_DATE_OVERRIDE` and exercises `predict` or `timeline` gets a partly-overridden result, which either produces a flaky test or, worse, a test that passes for the wrong reason. Second, `timestamp_detective` is the feature the README leads with, and its anchor is not pinnable, so its behaviour cannot be reproduced deterministically in a bug report.

## Precedent

#316 fixed exactly this for `exporter.py`'s numeric `--since` path. The fix was correct but scoped to that one call site, and the same class of bug was left in place elsewhere.

## Suggested fix

Audit the list above, replace `datetime.now()` with `date_utils.get_current_time()` wherever the value is a semantic "now" rather than a wall-clock measurement, and add a short comment at the sites that intentionally keep `datetime.now()` so the next audit does not have to re-derive the reasoning.

A lint rule or a test that greps for new `datetime.now()` occurrences outside an allowlist would keep it from regressing.

sanjana2505006:
## Summary
- Route semantic "now" through `date_utils.get_current_time()` in parser, predict, timeline, web, database, and timestamp_detective so `TERMSTORY_DATE_OVERRIDE` applies consistently
- Leave intentional wall-clock `datetime.now()` call sites in exporter, tui, and backup, with short comments explaining why
- Add an allowlist regression test so new bypasses don't slip back in

Fixes #407

## Test plan
- [x] `pytest tests/test_date_utils.py tests/test_timeline.py tests/test_predict.py tests/test_exporter.py tests/test_backup.py tests/test_timestamp_detective.py --timeout=60 -q`
- [ ] Set `TERMSTORY_DATE_OVERRIDE=2026-06-03 12:00:00` and confirm `termstory timeline` / `termstory predict` use that date
- [ ] Confirm backup filenames still use real wall-clock time

### DIFF
--- a/tests/test_date_utils.py
+++ b/tests/test_date_utils.py
@@ -1,7 +1,17 @@
 import pytest
 import os
+import re
+from pathlib import Path
 from datetime import datetime
 from termstory.date_utils import get_current_time, get_today_range, get_week_range, get_month_range, format_date_range
+
+# Call sites that intentionally use wall-clock time instead of get_current_time().
+_DATETIME_NOW_ALLOWLIST = {
+    "termstory/date_utils.py",  # implementation of get_current_time()
+    "termstory/exporter.py",   # Windows UTC-offset fallback
+    "termstory/tui.py",         # error log write time
+    "termstory/backup.py",      # unique backup filenames
+}
 
 def test_get_current_time(monkeypatch):
     # Without override
@@ -104,3 +114,21 @@ def test_parse_date_range_helper_chains_underlying_error():
     with pytest.raises(ValueError) as exc:
         parse_date_range_helper("not-a-real-date-at-all-zzzzz")
     assert exc.value.__cause__ is not None, "ValueError must chain the underlying parse error"
+
+def test_datetime_now_only_used_on_allowlisted_sites():
+    """Semantic 'now' must go through get_current_time() so TERMSTORY_DATE_OVERRIDE sticks."""
+    root = Path(__file__).resolve().parents[1] / "termstory"
+    offenders = []
+    for path in sorted(root.glob("*.py")):
+        rel = f"termstory/{path.name}"
+        if rel in _DATETIME_NOW_ALLOWLIST:
+            continue
+        text = path.read_text(encoding="utf-8")
+        for i, line in enumerate(text.splitlines(), 1):
+            if re.search(r"datetime\.now\s*\(", line):
+                offenders.append(f"{rel}:{i}: {line.strip()}")
+    assert not offenders, (
+        "datetime.now() outside the allowlist — use get_current_time() for semantic now, "
+        "or add the file to _DATETIME_NOW_ALLOWLIST with a comment at the call site:\n"
+        + "\n".join(offenders)
+    )
--- a/tests/test_date_utils.py
+++ b/tests/test_date_utils.py
@@ -119,16 +119,16 @@ def test_datetime_now_only_used_on_allowlisted_sites():
     """Semantic 'now' must go through get_current_time() so TERMSTORY_DATE_OVERRIDE sticks."""
     root = Path(__file__).resolve().parents[1] / "termstory"
     offenders = []
-    for path in sorted(root.glob("*.py")):
-        rel = f"termstory/{path.name}"
+    for path in sorted(root.rglob("*.py")):
+        rel = path.relative_to(root.parent).as_posix()
         if rel in _DATETIME_NOW_ALLOWLIST:
             continue
         text = path.read_text(encoding="utf-8")
         for i, line in enumerate(text.splitlines(), 1):
-            if re.search(r"datetime\.now\s*\(", line):
+            if re.search(r"\bdatetime\.(now|utcnow)\s*\(|\bdate\.today\s*\(", line):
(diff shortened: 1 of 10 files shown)


sanjana2505006:
@bitflicker64 please have a look and let me know if there's any scope for improvements :)

sanjana2505006:
run :miku /check-star

sanjana2505006:
:miku /review

bitflicker64:
@sanjana2505006 i'd recommend not using miku for review as the repo isnt indexed ( its using a older index )

bitflicker64:
if you are trying ecsoc this will give u extra points!!

tests/test_date_utils.py
```diff
@@ -104,3 +114,21 @@ def test_parse_date_range_helper_chains_underlying_error():
...
+    """Semantic 'now' must go through get_current_time() so TERMSTORY_DATE_OVERRIDE sticks."""
+    root = Path(__file__).resolve().parents[1] / "termstory"
+    offenders = []
+    for path in sorted(root.glob("*.py")):
+        rel = f"termstory/{path.name}"
+        if rel in _DATETIME_NOW_ALLOWLIST:
+            continue
+        text = path.read_text(encoding="utf-8")
+        for i, line in enumerate(text.splitlines(), 1):
+            if re.search(r"datetime\.now\s*\(", line):
```
bitflicker64: `datetime.utcnow()` and `date.today()` are also semantic "now" and currently slip through: `termstory/database.py:278` uses `datetime.utcnow()` today. Widening the pattern keeps those from creeping back too (you'd then need to either convert database.py:278 or allowlist it with a comment):

```suggestion
            if re.search(r"\bdatetime\.(now|utcnow)\s*\(|\bdate\.today\s*\(", line):
```

tests/test_date_utils.py
```diff
@@ -104,3 +114,21 @@ def test_parse_date_range_helper_chains_underlying_error():
     with pytest.raises(ValueError) as exc:
         parse_date_range_helper("not-a-real-date-at-all-zzzzz")
     assert exc.value.__cause__ is not None, "ValueError must chain the underlying parse error"
+
+def test_datetime_now_only_used_on_allowlisted_sites():
+    """Semantic 'now' must go through get_current_time() so TERMSTORY_DATE_OVERRIDE sticks."""
+    root = Path(__file__).resolve().parents[1] / "termstory"
+    offenders = []
+    for path in sorted(root.glob("*.py")):
+        rel = f"termstory/{path.name}"
```
bitflicker64: `glob("*.py")` only scans the top level, so a future subpackage would be invisible to this guard. `rglob` plus a root-relative path keeps the allowlist keys working:

```suggestion
    for path in sorted(root.rglob("*.py")):
        rel = path.relative_to(root.parent).as_posix()
```

bitflicker64 (commented):
The routing changes look good. Verified that `get_current_time()` returns a naive local datetime (same shape as `datetime.now()`), so the swaps in parser/predict/timeline/web/database/timestamp_detective are drop-in, and the removed `import time` in `web.py` has no remaining users in the file. The intentional wall-clock call sites and their comments make sense.

One gap worth closing: `termstory/database.py:278` uses `datetime.utcnow()`, a semantic "now" written into the DB, which both bypasses `TERMSTORY_DATE_OVERRIDE` and slips past the new regression test, since the regex only matches `datetime.now(`. (Note `datetime.utcnow().timestamp()` is also subtly wrong on non-UTC machines, since the naive UTC datetime gets interpreted as local time; switching it to `get_current_time()` would fix both, but since it changes stored timestamps it may deserve its own PR.) Inline suggestions below to at least widen the test's net.

PR #409, opened 2026-08-07
[Archive] Batch command inserts with executemany (fixes #406)

### LINKED ISSUE #406: [Enhancement] archive.py: per-row INSERT loop makes archiving large histories slow
## Summary

`archive_old_data()` copies rows one at a time. For each session it issues a SELECT, an INSERT, an optional FTS INSERT, another SELECT for commands, and then one INSERT plus one FTS INSERT **per command**. On a history with 100k archived commands that is upwards of 200k individual statement executions inside a single `BEGIN IMMEDIATE` transaction.

## Code

`termstory/archive.py`:

```python
for old_sess_id in session_ids:
    cursor.execute("SELECT ... FROM main.sessions WHERE id = ?", (old_sess_id,))
    ...
    cursor.execute("INSERT INTO archive.sessions ...")
    ...
    cursor.execute("SELECT ... FROM main.commands WHERE session_id = ?", (old_sess_id,))
    commands = cursor.fetchall()
    for cmd_ts, cmd_str, ... in commands:
        cursor.execute("INSERT INTO archive.commands ...")     # one per command
        if has_archive_fts5:
            cursor.execute("INSERT INTO archive.search_index ...")  # and another
```

The per-session structure exists for a real reason: `cursor.lastrowid` is needed to remap `session_id` foreign keys into the archive database, and the ID space can collide. That constraint is what makes the naive `INSERT INTO archive.x SELECT * FROM main.x` rewrite unsafe, and it should be preserved by any fix.

## Suggested approach

The session loop needs to stay serial to get each `lastrowid`, but the inner command loop does not. Within one session, all commands map to the same `new_sess_id`, so they can be batched:

```python
rows = [(cmd_ts, cmd_str, exit_code, new_sess_id, project_id_map.get(p), ...) for ... in commands]
cursor.executemany("INSERT INTO archive.commands (...) VALUES (?, ?, ?, ?, ?, ?, ?, ?)", rows)
```

with the same treatment for the paired `archive.search_index` inserts. That removes the large majority of the round trips.

A larger follow-up, if someone wants it: pre-compute the session ID mapping in one pass (reserve a contiguous ID range in the archive, or insert all sessions with `executemany` and read back the assigned ids), then batch commands across all sessions at once.

Worth adding a timing assertion to `tests/stress/test_massive_history_ingestion.py` so a regression here is visible.

sanjana2505006:
## Summary
- Batches per-session `commands` and FTS `search_index` inserts in `archive_old_data` using `executemany` instead of per-row `execute` calls, while keeping the outer session loop serial so `cursor.lastrowid` correctly remaps `new_sess_id` for FK integrity <cite repo="bitflicker64/Termstory" path="termstory/archive.py" start="188-206" end="188-206" />.
- Adds a stress test to catch any regression back to per-row inserts when archiving 10k commands.

## Changes
- `termstory/archive.py`: collect `cmd_rows` and `fts_rows` per session, then `cursor.executemany(...)` into `archive.commands` and `archive.search_index`, guarding the FTS batch with `if fts_rows:` <cite repo="bitflicker64/Termstory" path="termstory/archive.py" start="193-206" end="193-206" />.
- `stats["commands"]` now incremented by `len(cmd_rows)` instead of once per row, preserving the same stats semantics.
- The session-level loop over `session_ids` remains serial (`for old_sess_id in session_ids:`), since each session's `new_sess_id` comes from `cursor.lastrowid` after the `archive.sessions` insert and must stay fixed for every command in that session <cite repo="bitflicker64/Termstory" path="termstory/archive.py" start="157-180" end="157-180" />.
- `tests/stress/test_massive_history_ingestion.py`: new `test_archive_batched_inserts_timing` seeds 50 sessions × 200 commands (10k total) and asserts `archive_old_data` completes in under 30s, plus correctness of `stats["sessions"]` and `stats["commands"]`.

## Fixes
- Fixes #406 — per-row `INSERT` during archiving was slow for large histories.

## Testing
- `pytest tests/test_archive.py tests/test_archive_validation.py --timeout=60 -q`
- `pytest tests/stress/test_massive_history_ingestion.py::test_archive_batched_inserts_timing --timeout=60 -q`
- Manual spot-check recommended (per PR checklist): verify archived sessions still remap `session_id` correctly and FTS command rows land in the archive DB.

## Notes
- Automated review from `git-miku[bot]` confirmed the FK-remap logic is correct since `new_sess_id` is fixed per session and reused across batched `cmd_rows`.
- `coderabbitai[bot]` review was rate-limited and did not run; author notes the `git-miku` review index may be stale for current commits, so treat both as informational only, not a substitute for manual review.

### DIFF
--- a/termstory/archive.py
+++ b/termstory/archive.py
@@ -185,25 +185,35 @@ def archive_old_data(main_db_path: str, archive_db_path: str, days: int) -> Dict
                     VALUES (?, 'session_summary', ?, ?, ?)
                 """, (ai_summary, str(new_sess_id), new_proj_id, start_time))
 
-            # Fetch and insert commands
+            # Fetch and batch-insert commands (same new_sess_id for the whole session)
             cursor.execute("""
                 SELECT timestamp, command, exit_code, project_id, created_at, recovery_source, is_legacy
                 FROM main.commands WHERE session_id = ?
             """, (old_sess_id,))
             commands = cursor.fetchall()
-            for cmd_ts, cmd_str, exit_code, old_cmd_proj_id, cmd_ca, rec_src, is_leg in commands:
-                new_cmd_proj_id = project_id_map.get(old_cmd_proj_id)
-                cursor.execute("""
+            if commands:
+                cmd_rows = []
+                fts_rows = []
+                for cmd_ts, cmd_str, exit_code, old_cmd_proj_id, cmd_ca, rec_src, is_leg in commands:
+                    new_cmd_proj_id = project_id_map.get(old_cmd_proj_id)
+                    cmd_rows.append((
+                        cmd_ts, cmd_str, exit_code, new_sess_id,
+                        new_cmd_proj_id, cmd_ca, rec_src, is_leg,
+                    ))
+                    if has_archive_fts5:
+                        fts_rows.append((cmd_str, str(new_sess_id), new_cmd_proj_id, cmd_ts))
+
+                cursor.executemany("""
                     INSERT INTO archive.commands (timestamp, command, exit_code, session_id, project_id, created_at, recovery_source, is_legacy)
                     VALUES (?, ?, ?, ?, ?, ?, ?, ?)
-                """, (cmd_ts, cmd_str, exit_code, new_sess_id, new_cmd_proj_id, cmd_ca, rec_src, is_leg))
-                stats["commands"] += 1
+                """, cmd_rows)
+                stats["commands"] += len(cmd_rows)
 
-                if has_archive_fts5:
-                    cursor.execute("""
+                if fts_rows:
+                    cursor.executemany("""
                         INSERT INTO archive.search_index (content, type, ref_id, project_id, timestamp)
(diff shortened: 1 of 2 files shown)


sanjana2505006:
:miku /review

bitflicker64:
@sanjana2505006 miku review is stale isnt index for current commits i recommend not using it for now ..

bitflicker64 (approved):
## Code Review Summary

**Verdict: APPROVE**

No findings.

### Assessment
Per-session archive loop stays serial so `lastrowid` remaps session FKs correctly; `executemany` batches per-session `archive.commands` and FTS `search_index` inserts with the same parameter order, project_id remap, stats bookkeeping, and FTS gating as the previous per-row path. Stress test covers 10k commands with session/command count assertions.

### Issue counts
- P0: 0
- P1: 0
- P2: 0
- P3: 0

bitflicker64 (approved):

PR #411, opened 2026-08-08
[Timeline] Align fetch window with rendered days (fixes #404)

### LINKED ISSUE #404: [Bug] timeline.py: fetch window is one day wider than the render window, distorting bar scaling
## Summary

`render_timeline()` queries sessions starting from `now - days`, but only renders the `days` calendar dates from `now - (days - 1)` through today. Sessions on the extra oldest day are loaded, counted into `daily`, included in the `max_dur` used to scale every bar, and then never displayed.

## Code

`termstory/timeline.py`:

```python
start_ts = int((now - timedelta(days=days)).timestamp())
sessions = db.get_range_sessions(start_ts, int(now.timestamp()))
daily = _aggregate_sessions_by_day(sessions)

all_dates = [(now - timedelta(days=i)).strftime("%Y-%m-%d") for i in range(days - 1, -1, -1)]
for d in all_dates:
    daily.setdefault(d, 0)
max_dur = max(daily.values()) if daily else 0     # <- includes the invisible day
```

## Reproduce

```python
from datetime import datetime, timedelta
now, days = datetime.now(), 7
start = datetime.fromtimestamp(int((now - timedelta(days=days)).timestamp()))
rendered = [(now - timedelta(days=i)).strftime('%Y-%m-%d') for i in range(days-1,-1,-1)]
print('fetched from:', start.date())
print('rendered:', rendered[0], '..', rendered[-1])
```

Output:

```
fetched from: 2026-07-31
rendered:     2026-08-01 .. 2026-08-07
```

Sessions on 2026-07-31 are fetched but never rendered.

## Impact

If the invisible day happens to be the user's heaviest day in the window, every visible bar is scaled against a maximum that does not appear on screen. A day that is in fact the busiest of the seven shown renders as a short bar, with nothing to explain why. The effect is invisible when the extra day is quiet, which makes it an intermittent and confusing bug rather than an obviously wrong one.

## Suggested fix

Align the two windows. Either fetch from `now - timedelta(days=days - 1)` at start-of-day, or extend `all_dates` to `range(days, -1, -1)`. The first is preferable, since `days=7` reading as "the last 7 days shown" matches the flag's documentation. Clamping the query to start-of-day rather than the current wall-clock time would also make the earliest rendered bar complete rather than partial.

## Note

`stats.daily_activity_heatmap()` has the same off-by-one in its query window. It is harmless there today only because that function has no max-based scaling.

sanjana2505006:
## Summary
- Fetch timeline sessions from start-of-day of the oldest rendered date so the query matches the `days` shown
- Scale bars using only visible days, so a heavy day outside the window cannot shrink on-screen bars
- Add a regression test covering the off-by-one scaling case

Fixes #404

## Test plan
- [x] `pytest tests/test_timeline.py --timeout=60 -q`
- [ ] Run `termstory timeline --days 7` and confirm exactly 7 date rows appear
- [ ] Confirm the busiest visible day gets a full-width bar

### DIFF
--- a/termstory/timeline.py
+++ b/termstory/timeline.py
@@ -56,15 +56,15 @@ def render_timeline(db: Database, days: int = 30) -> str:
     if days <= 0:
         raise ValueError("days must be greater than 0")
     now = get_current_time()
-    start_ts = int((now - timedelta(days=days)).timestamp())
-    # Fetch sessions in range
+    # Match the rendered window: oldest shown day through today, from start-of-day.
+    start_day = datetime.combine((now - timedelta(days=days - 1)).date(), datetime.min.time())
+    start_ts = int(start_day.timestamp())
     sessions = db.get_range_sessions(start_ts, int(now.timestamp()))
     daily = _aggregate_sessions_by_day(sessions)
-    # Ensure all dates are represented
     all_dates = [(now - timedelta(days=i)).strftime("%Y-%m-%d") for i in range(days - 1, -1, -1)]
     for d in all_dates:
         daily.setdefault(d, 0)
-    max_dur = max(daily.values()) if daily else 0
+    max_dur = max(daily[d] for d in all_dates)
     lines: List[str] = []
     header = f"{'Date':<12} | Activity"
     separator = "-" * (len(header) + 2)
--- a/tests/test_timeline.py
+++ b/tests/test_timeline.py
@@ -83,7 +83,7 @@ def test_render_timeline_invalid_days(tmp_path):
         render_timeline(db, days=-5)
 
 
-def test_render_timeline_scales_only_visible_days(tmp_path):
+def test_render_timeline_scales_only_visible_days(tmp_path, monkeypatch):
     """A heavy day just outside the window must not shrink the visible bars."""
     from termstory.timeline import render_timeline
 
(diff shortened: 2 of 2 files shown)


bitflicker64 (commented):
Good catch on the window mismatch — the underlying bug is real. With the clock frozen at 08:30 I reproduced it on `main`: a heavy session on the day just outside the window pulled `max_dur` up and shrank today's bar from 40 blocks down to 4. Both halves of the fix are right, and scoping `max_dur` to `all_dates` is a nice belt-and-braces guard.

Two things before this can go in.

**1. Needs a rebase — #408 just landed and touches the same lines.**

`render_timeline` now opens with `now = get_current_time()` so `TERMSTORY_DATE_OVERRIDE` applies. Keep that line and layer your window change on top of it:

```python
    now = get_current_time()
    # Match the rendered window: oldest shown day through today, from start-of-day.
    start_day = datetime.combine((now - timedelta(days=days - 1)).date(), datetime.min.time())
    start_ts = int(start_day.timestamp())
```

Careful not to resolve the conflict back to `datetime.now()` — #408 added an allowlist test in `tests/test_date_utils.py` that fails if `timeline.py` reads the clock directly. I tried the resolution above locally and both `tests/test_timeline.py` and `tests/test_date_utils.py` pass.

**2. The new test only catches the bug in the morning.**

`outside` is planted at 12:00 on the day before the window, but the old `start_ts` was `now - timedelta(days=days)` — that same calendar day at the *current* time-of-day. So the session only falls inside the old fetch window when the wall clock is before 12:00:

| now | old `start_ts` | outside fetched by old code? |
|---|---|---|
| 08:30 | 08-05 08:30 | yes → test catches the bug |
| 14:30 | 08-05 14:30 | no → test passes on unfixed code |

CI ran at 20:07 UTC, so it went green vacuously. Worth freezing the clock instead of depending on when the suite runs — patch `termstory.timeline.datetime` (or `get_current_time`, post-rebase) with a fixed `now` and the assertion stops being time-of-day dependent.

Everything else looks good.

PR #413, opened 2026-08-09
[Config] Surface save_config write failures instead of swallowing the…

### LINKED ISSUE #402: [Bug] config.py: save_config swallows every write failure silently
## Summary

`save_config()` wraps its entire body in `try: ... except Exception: pass`. If the config cannot be written for any reason - read-only directory, full disk, permissions, a bad path from `XDG_CONFIG_HOME` - the function returns normally, logs nothing, and the caller has no way to tell the write did not happen.

## Code

`termstory/config.py`:

```python
def save_config(config: dict) -> None:
    config_path = get_config_path()
    dir_path = os.path.dirname(config_path)
    try:
        os.makedirs(dir_path, exist_ok=True)
        fd, tmp_path = tempfile.mkstemp(...)
        try:
            ...
            os.replace(tmp_path, config_path)
        except Exception:
            try:
                os.unlink(tmp_path)
            except Exception:
                pass
            raise
    except Exception:
        pass          # <- every failure, including the re-raise above, ends here
```

The inner `raise` is carefully written to clean up the temp file and propagate, and then the outer handler discards it.

## Reproduce

```python
import os, tempfile, termstory.config as C
d = tempfile.mkdtemp()
os.environ['XDG_CONFIG_HOME'] = d
os.environ['HOME'] = d
import importlib; importlib.reload(C)
p = C.get_config_path()
os.chmod(os.path.dirname(p), 0o555)      # read-only
C.save_config({'ai_enabled': True})
print('exists:', os.path.exists(p))
```

Actual output:

```
exists: False
```

No exception, no log line, no return value indicating failure.

## Impact

`termstory config set ...` prints its success message and exits 0 while nothing was persisted. The user changes their API key or provider, sees it confirmed, and the setting is gone on the next run with no explanation. `load_config()` also calls `save_config()` after migrating legacy keys, so a failed migration is silently repeated on every single invocation.

## Suggested fix

At minimum, `logger.exception(...)` in the outer handler so the failure is visible under `--verbose`. Better: let the exception propagate and have the CLI `config set` path catch it and print an actionable error. The atomic-write machinery is already correct, it just needs its failures to be audible.

## Related

While in this file: `load_config()` has an `except OSError` clause after an `except (json.JSONDecodeError, UnicodeDecodeError, OSError, ValueError, RecursionError)` clause. The second is unreachable, since `OSError` is already caught by the first.

sanjana2505006:
## Summary
- Stop swallowing every `save_config()` failure with a bare `except Exception: pass`
- Make `termstory config set` report a clear error and exit non-zero when the write fails
- Log migration/persist failures in `load_config()` without hiding them
- Remove the unreachable duplicate `except OSError` in `load_config()`

Fixes #402

## Test plan
- [x] `pytest tests/test_config.py tests/test_cli_commands.py::test_config_set_reports_save_failure --timeout=60 -q`
- [ ] Point config at a read-only directory and confirm `termstory config set ...` fails with `Failed to save config` (no green success line)
- [ ] Confirm a normal `termstory config set ai_enabled true` still succeeds and persists

### DIFF
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -1123,7 +1123,10 @@ def show_ui(
                 console.print("Continuing with legacy history fallback...")
         elif response in ("n", "no"):
             _cfg["has_seen_timestamp_prompt"] = True
-            save_config(_cfg)
+            try:
+                save_config(_cfg)
+            except OSError as e:
+                console.print(f"[bold red]Failed to save config:[/] {e}")
             console.print("Continuing with legacy history fallback...")
         else:
             console.print("Invalid response. Continuing with legacy history fallback...")
@@ -1552,8 +1555,15 @@ def config_set(key: str, value: str):
         converted_value = value
 
     set_config_value(config, key, converted_value)
-    save_config(config)
-    
+    try:
+        save_config(config)
+    except OSError as e:
+        Console(stderr=True).print(
+            f"[bold red]Failed to save config:[/] {e}\n"
+            "Check that the config directory is writable and has free disk space."
+        )
+        raise typer.Exit(code=1)
+
     set_val = get_config_value(config, key)
     from rich.markup import escape
     console.print(f"[green]Set config key '{escape(key)}' to '{escape(str(set_val))}'[/]")
--- a/termstory/config.py
+++ b/termstory/config.py
@@ -215,14 +215,6 @@ def load_config() -> dict:
             )
             config = {}
 
-        except OSError as e:
-            logger.warning(
-                "Could not read config file '%s': %s",
-                config_path,
-            e,
-            )
-            config = {}
-
     if not isinstance(config, dict):
         config = {}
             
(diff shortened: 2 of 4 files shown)


bitflicker64:
Light review — the change itself is right: swallowing every write failure with \`except Exception: pass\` was the bug, and surfacing it at the \`config set\` call site is the correct fix. Dropping the unreachable duplicate \`except OSError\` is a good catch too.

Blocking on CI though — tests fail on Python 3.10/3.11/3.12:

```
tests/test_cli_commands.py:489: in test_config_set_reports_save_failure
    runner = CliRunner(mix_stderr=True)
E   TypeError: CliRunner.__init__() got an unexpected keyword argument 'mix_stderr'
```

`mix_stderr` was removed in Click 8.2; stderr is merged into `result.output` by default now. 3.9 passes only because it resolves an older Click. Just use `CliRunner()`:

```suggestion
        runner = CliRunner()
```

One more thing to check while you're in there: `test_save_config_readonly_dir_raises` and `test_config_set_reports_save_failure` both rely on `chmod 0o555` to force the failure. That is a no-op for root, so they will silently pass without exercising anything in a root container. Worth a `pytest.mark.skipif(os.geteuid() == 0)` guard.

Ping me once CI is green and I'll merge.

PR #415, opened 2026-08-09
[Sanitizer] Stop FQDN redaction from eating versions and module paths…

### LINKED ISSUE #400: [Bug] sanitizer.py: FQDN_PATTERN destroys version pins, Docker tags, and branch names
## Summary

`FQDN_PATTERN` matches any dotted token and replaces it with `[REDACTED_HOST]` unless the last segment happens to be in the hardcoded `FILE_EXTENSIONS` set. Version numbers, image tags, branch names, and dotted directory names all match. The result is that a large share of ordinary commands are mangled before they reach the AI, the web report, or the TUI.

## Reproduce

```python
from termstory.sanitizer import redact_command
for t in [
    'pip install requests==2.31.0',
    'git checkout release-1.2.3',
    'docker run nginx:1.25.4-alpine',
    'python manage.py migrate --settings=app.settings.prod',
    'cd ~/Projects/my.app.dir',
]:
    print(repr(t), '->', repr(redact_command(t)))
```

Actual output on `main`:

```
'pip install requests==2.31.0'                          -> 'pip install requests==[REDACTED_HOST]'
'git checkout release-1.2.3'                            -> 'git checkout [REDACTED_HOST]'
'docker run nginx:1.25.4-alpine'                        -> 'docker run nginx:[REDACTED_HOST]'
'python manage.py migrate --settings=app.settings.prod' -> 'python manage.py migrate --settings=[REDACTED_HOST]'
'cd ~/Projects/my.app.dir'                              -> 'cd ~/Projects/[REDACTED_HOST]'
```

## Why this matters

None of these are hostnames and none are secrets. The damage is to the product's core value: the AI narrator is asked to summarise a day's work from commands where the package versions, branch names, and image tags have been blanked out, and the TUI and web report show the same holes. It also degrades search, since the redacted forms are what get indexed for the session summary path.

## Root cause

`FILE_EXTENSIONS` is a denylist-shaped allowlist. It only spares a token when the final dotted segment is a known file extension, so `2.31.0` (last segment `0`), `1.2.3` (`3`), and `settings.prod` (`prod`) all fall through to redaction.

## Suggested fix

Require a real TLD-shaped final segment before redacting: at minimum, the last segment should be alphabetic and 2 or more characters. That alone fixes the version-number cases. Beyond that, consider gating FQDN redaction on the token appearing in a network context (after `ssh`, `curl`, `scp`, `ping`, `://`, or `@`) rather than redacting every dotted token in every command.

Whatever the fix, the cases above make good regression tests.

sanjana2505006:
## Summary
- Require a TLD-shaped final segment (alphabetic, length ≥ 2) before FQDN redaction
- Only redact dotted tokens in a network context (`@`, `://`, or after ssh/curl/scp/ping/wget)
- Preserve version pins, Docker tags, branch names, module paths, and dotted directories
- Keep redacting real hosts in network commands
- Add regression tests for the issue #400 reproduce cases

Fixes #400

## Test plan
- [x] `pytest tests/test_sanitizer.py --timeout=60 -q`
- [ ] Confirm `pip install requests==2.31.0` and `docker run nginx:1.25.4-alpine` are unchanged
- [ ] Confirm `ping dev-server.local` and `curl https://api.internal.domain.local/...` still redact hosts

### DIFF
--- a/termstory/sanitizer.py
+++ b/termstory/sanitizer.py
@@ -111,6 +111,16 @@ FILE_EXTENSIONS = {
     'png', 'jpg', 'jpeg', 'gif', 'svg', 'zip', 'tar', 'gz', 'log', 'out'
 }
 
+# Host-shaped final labels. Used so we can redact real hostnames without a
+# command allowlist (which misses destination-arg forms like scp/rsync/psql -h).
+HOST_TLDS = {
+    'com', 'net', 'org', 'io', 'edu', 'gov', 'info', 'biz', 'co',
+    'app', 'dev', 'cloud', 'tech', 'me', 'ai', 'cc', 'xyz',
+    'us', 'uk', 'de', 'fr', 'jp', 'au', 'ca', 'in', 'br', 'cn', 'ru',
+    'local', 'internal', 'lan', 'corp', 'home', 'intranet', 'private',
+    'localdomain',
+}
+
 # Match standard FQDNs (e.g. host.domain.com)
 FQDN_PATTERN = re.compile(r'\b([a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)+)\b')
 
@@ -237,7 +247,7 @@ def redact_command(cmd: str) -> str:
     cmd = IP_ADDRESS_PATTERN.sub('[REDACTED_IP]', cmd)
     cmd = IPV6_ADDRESS_PATTERN.sub('[REDACTED_IP]', cmd)
     
-    # 6. FQDNs (excluding files, versions, and non-network dotted tokens)
+    # 6. FQDNs (excluding files, versions, and non-host dotted tokens)
     def fqdn_replacer(match):
         full_match = match.group(1)
         parts = full_match.split('.')
@@ -247,16 +257,13 @@ def redact_command(cmd: str) -> str:
         # Version pins / numeric tags (2.31.0, 1.25.4-alpine) aren't hostnames.
         if not ext.isalpha() or len(ext) < 2:
             return full_match
-        # Only redact in a network context — otherwise branch names, module
-        # paths, and dirs like my.app.dir get blanked for no privacy gain.
         prefix = cmd[: match.start(1)]
+        # user@host and scheme://host are always host-shaped.
         if prefix.endswith("@") or prefix.endswith("://"):
             return "[REDACTED_HOST]"
-        if re.search(
-            r"(?:^|[\s;|&])(?:ssh|curl|scp|ping|wget)(?:\s+-\S+)*\s+$",
-            prefix,
-            re.IGNORECASE,
-        ):
+        # Otherwise only redact when the last label looks like a real TLD.
+        # Keeps module paths (app.settings.prod) and dirs (my.app.dir) intact.
+        if ext in HOST_TLDS:
             return "[REDACTED_HOST]"
         return full_match
 
--- a/termstory/sanitizer.py
+++ b/termstory/sanitizer.py
@@ -237,16 +237,29 @@ def redact_command(cmd: str) -> str:
     cmd = IP_ADDRESS_PATTERN.sub('[REDACTED_IP]', cmd)
     cmd = IPV6_ADDRESS_PATTERN.sub('[REDACTED_IP]', cmd)
     
-    # 6. FQDNs (excluding files)
+    # 6. FQDNs (excluding files, versions, and non-network dotted tokens)
     def fqdn_replacer(match):
         full_match = match.group(1)
         parts = full_match.split('.')
         ext = parts[-1].lower()
         if ext in FILE_EXTENSIONS:
-            # Keep as-is, looks like a source/config file
             return full_match
-        return '[REDACTED_HOST]'
-        
+        # Version pins / numeric tags (2.31.0, 1.25.4-alpine) aren't hostnames.
+        if not ext.isalpha() or len(ext) < 2:
+            return full_match
+        # Only redact in a network context — otherwise branch names, module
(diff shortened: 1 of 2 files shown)


bitflicker64:
Light review. The over-redaction problem in #400 is real and the TLD-shape check (`ext.isalpha() and len(ext) >= 2`) is a clean fix for version pins and Docker tags — no complaints there.

The part I can't merge as-is is the second gate: requiring a network context via a five-command allowlist (`ssh|curl|scp|ping|wget`). This is a privacy tool, so trading over-redaction for under-redaction is the worse direction, and the allowlist leaks real internal hostnames today. Checked out your branch and ran it:

```
'scp report.txt build.internal.corp:/tmp'      -> unchanged
'psql -h prod-db.internal.corp -U admin'       -> unchanged
'rsync -a ./ backup.internal.corp:/data'       -> unchanged
'mysql -h db.internal.corp'                    -> unchanged
'ssh admin@dev.internal.corp'                  -> ssh admin@[REDACTED_HOST]   (ok)
'ping dev-server.local'                        -> ping [REDACTED_HOST]        (ok)
```

Note `scp` is already on your allowlist and still leaks — the regex is `(?:ssh|curl|scp|ping|wget)(?:\s+-\S+)*\s+$`, which only matches when the host is the token right after the command and its flags. For `scp`, the host is usually the *destination* arg, so the prefix ends in `report.txt ` and the match fails. Same shape breaks `ssh -o SomeOpt=x host.internal` and any command where the host isn't in first position.

Widening the allowlist doesn't really fix this — it's unbounded (`psql`, `mysql`, `rsync`, `telnet`, `nc`, `sftp`, `dig`, `host`, `nslookup`, `kubectl`, `redis-cli`, ...) and position-sensitive on top of that.

Suggestion: drop the command allowlist and lean on the token itself. Keep `@` / `://` as strong signals, then decide from the TLD:

- redact when the last segment is a known/host-shaped TLD (`com`, `net`, `org`, `io`, `local`, `internal`, `lan`, `corp`, plus a public-suffix-ish set)
- keep when it's in `FILE_EXTENSIONS`, or the token looks like a module path / version / branch (already covered by the `isalpha` check)

That keeps every case in your new test green while still catching `psql -h prod-db.internal.corp`. `app.settings.prod` and `my.app.dir` stay safe because `prod`/`dir` aren't TLDs.

If you'd rather keep this scoped tight, splitting it works too: land the `isalpha`/length TLD-shape check on its own (that alone closes the version-pin and Docker-tag complaints in #400) and handle the network-context question in a follow-up. Happy to merge that smaller version right away.

bitflicker64:
Thanks for the rework — dropping the command allowlist for `HOST_TLDS` + destination shapes is the right structure, and all four leaks from my last pass are now closed (`scp`/`rsync` dest args, `psql -h`, `mysql -h`). Checking `@` / `://` before `FILE_EXTENSIONS` so `curl https://service.py` still redacts is a nice touch.

Two new over-redactions crept in, both because the `-h` and `host:/path` checks now run *before* `FILE_EXTENSIONS`. Ran your branch:

```
'du -h report.txt'                                -> 'du -h [REDACTED_HOST]'
'sort -h data.log'                                -> 'sort -h [REDACTED_HOST]'
'tar -h archive.tar.gz'                           -> 'tar -h [REDACTED_HOST]'
'docker run -v ./nginx.conf:/etc/nginx/nginx.conf' -> 'docker run -v ./[REDACTED_HOST]:/etc/nginx/nginx.conf'
'docker run -v /opt/site.conf:/etc/site.conf'      -> 'docker run -v /opt/[REDACTED_HOST]:/etc/site.conf'
```

1. **`-h` is "human readable" far more often than "host"** — `du -h`, `df -h`, `sort -h`, `tar -h`, `ls -h`. Note `tar -h archive.tar.gz` gets redacted even though `gz` is in `FILE_EXTENSIONS`, because the `-h` branch returns first.
2. **`token:/path` matches Docker volume mounts**, not just `scp` destinations.

Both fall out from ordering plus one extra anchor. Move the `FILE_EXTENSIONS` check back above the shape checks (keeping it below `@` / `://`, which you were right to prioritise), and require the `host:` token to start its argument:

```suggestion
        if ext in FILE_EXTENSIONS:
            return full_match
        # Destination/arg shapes: -h host, --host=host, host:/path, host:port.
        # Redact regardless of TLD (covers build.example.museum:/tmp).
        if re.search(r'(?:^|[\s])(-h|--host)(\s+|=)$', prefix):
            return "[REDACTED_HOST]"
        # Only when the token starts the arg — else ./nginx.conf:/etc/... matches.
        if re.match(r':(\d+|/)', suffix) and re.search(r'(?:^|[\s=])$', prefix):
            return "[REDACTED_HOST]"
```

A real `scp` destination is preceded by a space, so the anchor keeps it; `./nginx.conf` and `/opt/site.conf` are preceded by `/`, so they're dropped. I applied exactly this locally and confirmed all five regressions above go away, every assertion in your new `test_fqdn_redacts_hosts_without_command_allowlist` still holds (including the `.museum` and `service.py` cases), and `tests/test_sanitizer.py` is 24/24 green.

Worth adding `conf` to `FILE_EXTENSIONS` while you're here, and regression tests for `du -h <file>` and a `-v` volume mount so this ordering doesn't silently flip back later.

One thing I'd flag but won't block on: `HOST_TLDS` is a fixed list, so bare hosts on TLDs outside it (`box.internal.museum`, `host.se`, `db.cn.example`) stay unredacted unless they hit `@`, `://`, `-h`, or `host:`. That's an acceptable trade for this PR — the strong-context paths cover the common cases — but it may be worth a follow-up issue.

Fix those two and I'll merge.

PR #419, opened 2026-08-12
fix: redact commands and commits in exports

### LINKED ISSUE #398: [Security] exporter.py / notebook.py: `export` and `notebook` write raw unredacted commands to disk
## Summary

`termstory export` (JSON and CSV) and `termstory notebook` write the raw contents of the `commands.command` column straight to the output file. Neither module imports the sanitizer, so credentials that TermStory redacts everywhere else are written verbatim into a file the user is likely to attach to a ticket, paste into a chat, or commit.

## Evidence

Every other consumer of command text redacts. `exporter.py` and `notebook.py` are the only ones that do not:

```
$ grep -rn "redact_command\|sanitize_session_commands" termstory/*.py | grep -v sanitizer.py
termstory/agy.py:...
termstory/ai.py:...
termstory/ask.py:...
termstory/cli.py:794
termstory/formatter.py:191,202,447,458
termstory/web.py:96
```

`exporter.py` and `notebook.py` do not appear.

The write paths are:

- `exporter.serialize_sessions_to_dict()` - `"command": cmd.command`
- `exporter.export_csv()` - `"command_text": cmd.command`
- `notebook.generate_notebook()` - `lines.append(f"    {cmd.command}")`

Commit messages are copied raw in the same functions (`message`, `cleaned_message`).

## Why this matters

This is the same class of bug as #346, where `report.html` embedded raw credential-bearing commands. That was fixed for `web.py` by routing through `redact_command()` at `web.py:96`, but the export and notebook paths were never given the same treatment. An export is arguably a higher-risk surface than the HTML report, since exports exist specifically to be moved off the machine.

## Reproduce

1. Ensure history contains something like `export AWS_SECRET_ACCESS_KEY=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY`.
2. `termstory export --format json -o out.json`
3. `grep -i secret out.json` returns the key in cleartext.

The same command rendered through `termstory web` or an AI summary comes out as `[REDACTED]`.

## Suggested fix

Route command and commit text through `sanitizer.redact_command()` in `serialize_sessions_to_dict()`, `export_csv()`, and `generate_notebook()`. Consider applying the `should_blacklist_command()` session gate too, matching `sanitize_session_commands()`. If a raw export is a legitimate use case, put it behind an explicit `--no-redact` opt-in flag rather than making it the default.

deepmhatre13:
Closes #398
## Summary

- Redact command text in JSON and CSV exports using the existing sanitizer.
- Redact command text in generated notebooks.
- Redact commit `message` and `cleaned_message` before writing them to export or notebook output.
- Apply the existing session-level blacklist behavior to prevent security/authentication commands from being written in raw form.
- Add regression tests covering JSON, CSV, notebook, and blacklisted-session output.

### `termstory/exporter.py`

- Updated `serialize_sessions_to_dict()` to use `sanitize_session_commands()` for command redaction and session blacklist handling.
- Updated commit `message` and `cleaned_message` to use `redact_command()`.
- Updated `export_csv()` so both command text and serialized commit messages are sanitized.
- Prevented `cleaned_message` from bypassing the sanitizer in CSV output.

### `termstory/notebook.py`

- Updated `generate_notebook()` to redact command text before writing it to the generated notebook.
- Added session-level blacklist handling for security/authentication commands.
- Redacted commit messages before adding them to the notebook.

## Tests

Added regression coverage for:

- JSON command secret redaction.
- JSON commit message redaction.
- CSV command and commit redaction.
- JSON blacklisted-session handling.
- CSV blacklisted-session handling.
- Notebook command redaction.
- Notebook blacklisted-session handling.
- Notebook commit message redaction.

## Validation

```text
50 passed
57 passed
35 passed
43 passed
git diff --check — clean
```

The full test suite still has pre-existing environment/platform-specific collection failures involving missing `textual`, Windows `os.geteuid()` compatibility, and Windows file-locking behavior.

### DIFF
--- a/termstory/exporter.py
+++ b/termstory/exporter.py
@@ -10,8 +10,15 @@ from dateutil import parser as date_parser
 from termstory.database import Database
 from termstory.date_utils import get_current_time
 from termstory.models import Session, Command, Project
+from termstory.sanitizer import redact_command, sanitize_session_commands
 
 _FAR_FUTURE_TS: int = 9_999_999_999 # ~year 2286, safely beyond any real session timestamp
+
+# Placeholder rendered in place of commands when a session is fully blacklisted
+# (e.g. ``vault`` / ``aws configure`` / ``gh auth`` operations). Reuses the
+# semantics documented in CONTRIBUTING.md so that raw security/authentication
+# command text can never reach disk via the export paths.
+_BLACKLISTED_SESSION_MARKER = "[REDACTED: Security/Authentication Operations]"
 
 def _get_timestamp(dt: datetime) -> int:
     """Safely get Unix timestamp from datetime object on all platforms, including Windows."""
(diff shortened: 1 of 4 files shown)


deepmhatre13:
:miku /ecsoc

PR #420, opened 2026-08-12
fix: sanitize MCP snapshot environment data

### LINKED ISSUE #399: [Security] mcp_snapshot.py: IDE environment variable values are persisted without redaction
## Summary

`capture_ide_state()` scans the whole process environment and stores the **values** of any variable whose name contains `VSCODE_`, `IDEA_`, `JETBRAINS_`, `XCODE_`, `NVIM_`, or `CURSOR_`, plus `TERM_PROGRAM`, `EDITOR`, and `VISUAL`. Those values go into the `mcp_snapshots.payload` JSON column untouched. The sanitizer is never involved.

## Code

`termstory/mcp_snapshot.py`:

```python
for k, v in os.environ.items():
    k_upper = k.upper()
    if any(term in k_upper for term in ["VSCODE_", "IDEA_", "JETBRAINS_", "XCODE_", "NVIM_", "CURSOR_"]):
        ide_vars[k] = v          # <- raw value, stored verbatim
```

and the payload is written as-is:

```python
db.save_mcp_snapshot(session_id=session_id, source="cli", payload=snapshot, ...)
```

`grep -rn "redact_command" termstory/mcp_snapshot.py` returns nothing.

## Why this matters

Prefix matching on the variable *name* is a poor proxy for "safe to store". Real variables that match and routinely carry sensitive values include:

- `VSCODE_GIT_ASKPASS_*` and `VSCODE_IPC_HOOK_CLI` - local socket and helper paths that leak the username and session id
- `NVIM_LISTEN_ADDRESS` - socket path containing the username
- `JETBRAINS_*` license and toolbox variables

Beyond the prefix set, `capture_git_status()` stores the full `git status --porcelain` output, which is a list of the user's uncommitted file paths, also unredacted.

DATA_PRIVACY.md tells users that data leaving the local machine is redacted. Snapshots are stored in the same SQLite database that `termstory export` reads and that users share when filing bugs, so "local only" is not a sufficient defence here.

## Suggested fix

- Run every captured env value through `sanitizer.redact_command()` before it reaches the payload.
- Switch to an allowlist of variable names actually needed for IDE detection, rather than the current substring prefix scan. Detection only needs to know *which* IDE, not the full value.
- Consider storing just `ide_name` plus a boolean set of detected markers, dropping the values entirely.

deepmhatre13:
Closes #399
## Summary

- Stop persisting raw IDE-prefixed environment variable values in MCP snapshots.
- Preserve IDE detection using safe signals while keeping sensitive environment values out of `env_vars`.
- Sanitize uncommitted Git file paths before they are persisted.
- Reuse the existing `redact_command()` sanitizer.
- Add regression tests covering sensitive IDE environment values, IDE detection, Git paths, and persisted snapshot payloads.

## Testing

- `python -m pytest tests/test_mcp_snapshot.py -q` — 13 passed
- `tests/test_replay.py` + `tests/test_sanitizer.py` — 35 passed
- `tests/test_mcp_snapshot.py` + `tests/test_database.py` — 28 passed
- `git diff --check` — clean

The CLI test collection has a pre-existing Windows-specific failure because `os.geteuid()` is unavailable on Windows; this is unrelated to the changes in this PR.

## Validation

A smoke test against the real environment confirmed VS Code detection still works while `env_vars` remains empty, preventing raw IDE environment values from being persisted.

### DIFF
--- a/termstory/mcp_snapshot.py
+++ b/termstory/mcp_snapshot.py
@@ -1,31 +1,54 @@
+import getpass
 import logging
 import os
+import re
 import sqlite3
 import subprocess
 import time
 from typing import Dict, Any, List
 
+from termstory.sanitizer import redact_command
+
 logger = logging.getLogger(__name__)
 
+# IDE/editor environment variable prefixes. Their *presence* drives IDE
+# detection, but their values are never persisted: they routinely hold socket
+# paths (e.g. VSCODE_IPC_HOOK_CLI, NVIM_LISTEN_ADDRESS), askpass helpers,
+# session identifiers, or license/locally-identifying data.
+IDE_PREFIXES = ("VSCODE_", "IDEA_", "JETBRAINS_", "XCODE_", "NVIM_", "CURSOR_")
+
+# General editor detection signals. TERM_PROGRAM is a program name (safe to
+# persist verbatim); EDITOR/VISUAL normally name an editor binary, so only
+# the trailing command name is retained (no home/username path component).
+GENERAL_EDITOR_SIGNALS = ("EDITOR", "VISUAL")
+
 def capture_ide_state() -> Dict[str, Any]:
-    """Capture active IDE state from environment variables"""
-    ide_vars = {}
+    """Capture active IDE state from environment variables.
+
+    IDE detection relies on the *presence* of IDE-prefixed environment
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

PR #421, opened 2026-08-12
fix: prevent backup filename collisions

### LINKED ISSUE #416: [Bug] backup.py: rapid consecutive backups can overwrite each other
## Description

`backup_db()` generates backup filenames using timestamps with only second-level precision. When two backups are created within the same second, both operations can produce the same filename.

For example:

```text
termstory_backup_20260812_073800.db
termstory_backup_20260812_073800.db
```

The second backup can therefore overwrite the first backup, reducing the number of independent recovery points available to the user.

This can occur when automated and manual backups run close together or when multiple backup requests are triggered in quick succession.

## Expected Behavior

Every successful backup operation should create a unique backup file, even when multiple backups occur within the same second.

## Suggested Fix

Use a collision-resistant filename strategy, such as microsecond precision or another atomic/unique naming mechanism.

Add a regression test that performs two backup operations immediately and verifies that they produce two distinct backup files.

deepmhatre13:
Closes #416
## Summary

- Add microsecond precision to backup filenames to prevent rapid consecutive backups from overwriting each other.
- Preserve lexicographic ordering so existing backup rotation continues to remove the oldest backups first.
- Add a regression test verifying that two immediate backup calls create distinct valid SQLite files.
- Strengthen backup rotation coverage to verify oldest-first deletion.

## Implementation

`backup_db()` now generates filenames using:

```text
termstory_backup_YYYYMMDD_HHMMSS_microseconds.db
```

The fixed-width microsecond component preserves chronological ordering while allowing multiple backups within the same second to have distinct filenames.

The implementation continues to use `datetime.now()`, preserving compatibility with the existing `MockDatetime` tests.

## Tests

- `python -m pytest tests/test_backup.py -q` — 4 passed
- `python -m pytest tests/test_cli_restore_confirmation.py -q` — 9 passed
- `python -m pytest tests/test_database.py tests/test_database_queries.py tests/test_archive.py tests/test_corrupt_db.py tests/test_corrupt_db2.py tests/test_corrupt_db3.py tests/test_save_data_none_safety.py tests/test_fts5.py tests/test_fts5_new.py -q` — 32 passed
- `git diff --check` — clean

The full test suite has pre-existing Windows/environment-specific collection failures involving `textual`, `os.geteuid`/`os.getuid`, and Windows process/file-lock behavior. These are unrelated to this change.

## Additional Verification

- `restore_db()` remains unchanged.
- The CLI continues to display the path returned by `backup_db()`.
- No database schema changes were required.
- Backup rotation continues to retain a maximum of 10 backups and removes the oldest backups first.

### DIFF
--- a/termstory/backup.py
+++ b/termstory/backup.py
@@ -20,6 +20,11 @@ def _get_backup_dir() -> str:
 def backup_db() -> str:
     """Create a timestamped backup of the TermStory database.
 
+    Backup filenames use a microsecond-precision timestamp
+    (``termstory_backup_YYYYMMDD_HHMMSS_mmmmmm.db``) so back-to-back backups
+    never collide. The fixed-width timestamp keeps filenames lexicographically
+    sortable, which backup rotation relies on to find the oldest backups.
+
     Returns:
         The absolute path to the created backup file.
     """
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

deepmhatre13:
sentinel ecsoc bot ? :miku

bitflicker64:
> sentinel ecsoc bot ? :miku

bot must be down take it to ecsoc admins

PR #422, opened 2026-08-12
fix: prevent stale project root cache results

### LINKED ISSUE #417: [Bug] project.py: cached project roots become stale after filesystem changes
## Description

`_find_project_root_cached()` caches project-root detection results based only on the input path. However, project-root detection depends on filesystem state, such as the presence of `.git` or other project markers.

If the filesystem changes after a path has been cached, subsequent calls can continue returning the old project root.

For example:

1. A directory is inspected before it contains a `.git` directory.
2. The project root is cached.
3. A repository is initialized or a project marker is created inside a subdirectory.
4. The same path is queried again.
5. The cached result is returned instead of detecting the new project root.

This can cause commands to be attributed to the wrong project.

## Expected Behavior

Project-root detection should reflect relevant filesystem changes rather than indefinitely returning a stale cached result.

## Suggested Fix

Review the caching strategy for filesystem-dependent project detection.

Possible approaches include:

- Avoiding caching for mutable filesystem state.
- Using a bounded TTL.
- Providing an appropriate cache invalidation strategy.

Add regression tests covering project-marker creation and removal after an initial project-root lookup.

deepmhatre13:
Closes #417
## Summary

- Replace the permanent `lru_cache` used for project-root detection with a bounded TTL-based LRU cache.
- Allow filesystem changes to project markers to be reflected after cache expiry.
- Preserve the existing 1024-entry cache limit and LRU behavior.
- Add a configurable `project_root_cache_ttl` setting with a 60-second default.
- Add regression tests for marker creation, marker removal, cache hits, and TTL/config behavior.
- Document the new configuration option.

## Implementation

`find_project_root()` keeps its existing public API.

The project-root cache now stores the cached result together with its timestamp and refreshes the result when the configured TTL expires.

The cache:

- Uses a 60-second default TTL.
- Preserves the existing 1024-entry maximum.
- Maintains LRU ordering.
- Uses the existing `datetime`/TTL-style caching approach used by the project.
- Avoids performing a filesystem walk on every lookup.

The existing network-mount/NFS timeout caching mechanism was left unchanged.

## Configuration

Added:

```text
project_root_cache_ttl = 60
```

The setting is registered in the configuration defaults and CLI validation and documented in `docs/configuration.md`.

## Tests

- Added regression coverage for project-marker creation after an initial lookup.
- Added regression coverage for project-marker removal after an initial lookup.
- Added coverage confirming cached results are reused within the TTL.
- Added configuration/default/clamping/fallback tests.
- `test_project.py` — 22 passed across repeated runs.
- `py_compile` — passed.
- Ruff — no new findings.
- `git diff --check` — clean.

## Known Environment Issues

The following failures are pre-existing and unrelated to this change:

- Windows symlink privilege failure in `test_find_project_root_symlink_escape`.
- Windows path mismatch in `test_detect_projects`.
- `os.geteuid()` collection failure in `tests/test_config.py` on Windows.

These were left unchanged because they are outside the scope of this issue.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -42,5 +42,6 @@ Key configuration parameters:
 - `request_timeout_seconds` (int): HTTP request timeout (in seconds) for LLM API calls. Defaults to `30`.
 - `max_query_log` (int): Maximum number of captured database query profiler entries before older entries are trimmed. Defaults to `10000`.
 - `providers.<name>.<param>`: Provider-specific endpoints, API keys, and model names.
+- `project_root_cache_ttl` (int, seconds): How long resolved project-root results are cached before being re-checked against the filesystem. A shorter value detects newly created/removed project markers (`.git`, `package.json`, ...) sooner but at the cost of more frequent filesystem walks. Defaults to `60`.
 
 ---
--- a/termstory/cli.py
+++ b/termstory/cli.py
@@ -1505,6 +1505,7 @@ def config_set(key: str, value: str):
         "reminder_poll_interval": 300,
         "clustering_threshold": 0.6,
         "nfs_timeout_cache_ttl": 60,
+        "project_root_cache_ttl": 60,
     }
 
     default_val = KNOWN_DEFAULTS.get(effective_key)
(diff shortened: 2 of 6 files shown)


deepmhatre13:
:miku /ecsoc

PR #424, opened 2026-08-12
fix: prevent concurrent sleep daemon startup

### LINKED ISSUE #418: [Bug] reminder.py: concurrent daemon startup can create duplicate sleep daemons
## Description

The sleep daemon startup logic checks whether the PID file exists before starting the daemon. The check and process creation are separate operations.

This creates a TOCTOU race when two processes attempt to start the daemon at approximately the same time:

1. Process A checks for the PID file and finds none.
2. Process B checks for the PID file and also finds none.
3. Process A starts a daemon.
4. Process B starts another daemon.

Both processes can therefore start their own sleep daemon.

Duplicate daemons can result in duplicate consolidation work, unnecessary resource usage, competing database operations, and inconsistent PID-file state.

## Expected Behavior

Only one sleep daemon should be able to acquire ownership and run at a time, including when startup is triggered concurrently.

## Suggested Fix

Replace the non-atomic "check PID file then spawn" sequence with an atomic locking or ownership mechanism.

Add a regression test that attempts concurrent daemon startup and verifies that only one daemon is created.

deepmhatre13:
Closes #418
## Summary

- Replace the non-atomic PID-file check-then-spawn sequence with atomic PID-file ownership acquisition.
- Prevent concurrent TermStory invocations from spawning duplicate sleep daemons.
- Preserve stale PID recovery and existing daemon PID/cleanup behavior.
- Release the ownership claim when placeholder creation or daemon startup fails.
- Add regression coverage for normal startup, running-daemon detection, stale PID recovery, and concurrent startup.

## Implementation

`start_sleep_daemon()` now uses:

```python
os.open(pid_file, os.O_CREAT | os.O_EXCL | os.O_WRONLY)
```

to atomically acquire ownership of the daemon PID file.

The winning caller writes a temporary live-owner PID before spawning the daemon. Existing `run_sleep_daemon()` behavior remains responsible for writing the actual daemon PID.

When another caller encounters an existing PID file, it checks whether the recorded process is alive. Live ownership causes the caller to defer rather than spawn another daemon.

Stale or invalid PID files can be reclaimed, allowing future daemon startup to proceed.

If placeholder creation or `Popen()` fails, the ownership claim is cleaned up so it does not permanently block future startup.

## Tests

Added coverage for:

- Starting when no daemon is running.
- Deferring when a live daemon PID exists.
- Reclaiming a stale PID file.
- Eight concurrent startup attempts resulting in exactly one daemon spawn.

Validation:

- New daemon tests — passed.
- Existing daemon/reminder tests — passed.
- `tests/test_sleep.py` — 4 passed.
- Ruff — no new findings.
- `git diff --check` — clean.

Some unrelated environment issues remain around RAG-stack imports and existing signal-handler/pytest-timeout behavior on this Windows environment.

### DIFF
--- a/termstory/reminder.py
+++ b/termstory/reminder.py
@@ -402,20 +402,62 @@ def consolidate_sleep_contexts(db, force: bool = False) -> int:
 
 
 def start_sleep_daemon(db_path: str):
-    """Spawns the sleep daemon in the background if it's not already running."""
+    """Spawns the sleep daemon in the background if it's not already running.
+
+    Daemon ownership is claimed atomically: the PID file is created with
+    ``O_CREAT | O_EXCL`` so that, of any number of concurrent invocations,
+    exactly one wins the race and goes on to spawn the daemon. Every other
+    invocation finds the file already exists and defers to the running (or
+    just-started) daemon, which later overwrites the placeholder PID with its
+    own. A stale PID file (a daemon that died without cleaning up) is reclaimed
+    so the daemon can be restarted.
+    """
     import sys
     import subprocess
-    
+
     pid_file = os.path.join(get_app_dir("data"), "sleep_daemon.pid")
-    if os.path.exists(pid_file):
+    os.makedirs(os.path.dirname(pid_file), exist_ok=True)
+
+    for attempt in range(2):
         try:
-            with open(pid_file, "r") as f:
-                pid = int(f.read().strip())
-            os.kill(pid, 0)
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

bitflicker64 (approved):

deepmhatre13:
@bitflicker64 merge?

PR #435, opened 2026-08-15
perf(database): eliminate project stats N+1 queries

### LINKED ISSUE #423: [Performance] database.py: get_projects_by_ids() performs N+1 session queries
## Description

`get_projects_by_ids()` retrieves the requested projects and then performs a separate session aggregate query for each project to calculate session counts and duration statistics.

This creates an N+1 query pattern:

```text
1 query to retrieve projects
+
N queries to retrieve session aggregates
```

For a large number of projects, this can result in hundreds or thousands of database queries for a single operation.

The issue becomes particularly relevant for exporters, reports, notebook generation, and other consumers that may request many projects simultaneously.

## Expected Behavior

Project metadata and session aggregate statistics should be retrieved with a bounded number of database queries regardless of the number of projects requested.

## Suggested Fix

Replace the per-project aggregate queries with a single aggregate query using `GROUP BY`, a `LEFT JOIN`, or another bulk-fetch strategy.

The resulting aggregates can then be mapped back to the requested projects.

Add a regression/performance test verifying that retrieving multiple projects does not execute one session query per project.

deepmhatre13:
Closes #423
## Summary

Eliminates the N+1 query pattern in `Database.get_projects_by_ids()`.

Previously, retrieving multiple projects executed one project query followed by a separate session aggregate query for every project. The implementation now retrieves project metadata and session statistics using a single bulk `LEFT JOIN` + `GROUP BY` query.

## Changes

- Replaced per-project session aggregate queries with a single bulk query.
- Added `LEFT JOIN` to preserve projects with zero sessions.
- Uses `COUNT(s.id)` so zero-session projects correctly return `session_count = 0`.
- Preserves `total_time = 0` for projects without sessions.
- Preserves the empty `project_ids` fast path.
- Added deterministic `ORDER BY p.id`.
- Added correctness and performance regression tests.
- No schema migrations or API changes.

## Tests

Full database test file:

```text
python -m pytest tests/test_database.py -v --tb=short

### DIFF
--- a/termstory/database.py
+++ b/termstory/database.py
@@ -860,23 +860,21 @@ class Database:
             
             placeholders = ",".join("?" for _ in project_ids)
             cursor.execute(f"""
-                SELECT id, name, path, first_seen, last_seen, project_context
-                FROM projects
-                WHERE id IN ({placeholders})
+                SELECT p.id, p.name, p.path, p.first_seen, p.last_seen,
+                       COUNT(s.id) AS session_count,
+                       SUM(s.duration_seconds) AS total_time,
+                       p.project_context
+                FROM projects p
+                LEFT JOIN sessions s ON p.id = s.project_id
+                WHERE p.id IN ({placeholders})
+                GROUP BY p.id
+                ORDER BY p.id
             """, project_ids)
             
             rows = cursor.fetchall()
             projects = []
             for row in rows:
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

PR #441, opened 2026-08-18
fix: make backup rotation resilient to clock skew

### LINKED ISSUE #429: fix: backup rotation deletes newest instead of oldest when clock skews
## Bug Report

**Describe the bug**
Backup rotation uses lexicographic sorting of filenames to identify the oldest backups for deletion. If the system clock skews backward (e.g. NTP correction, VM snapshot restore), a newer backup can have an earlier timestamp and get deleted instead of the actual oldest backup.

**Expected behavior**
Rotation should use file creation time (`os.path.getctime()`) as a tiebreaker, not rely solely on filename sorting.

**Impact**
Low — only affects systems with clock skew, but could cause data loss of recent backups.

deepmhatre13:
Closes #429
## Summary

Fixes backup rotation incorrectly deleting newer backups when the system clock moves backward.

Previously, `backup_db()` sorted backup filenames lexicographically. Since the filename contains a wall-clock timestamp, a backward clock adjustment could cause a newer backup to receive an earlier filename and be incorrectly selected for deletion.

The rotation logic now orders backups using `os.path.getctime()` instead.

## Changes

* Changed backup rotation ordering from filename lexicographic sorting to filesystem timestamp ordering.
* Preserved the existing maximum retention limit of 10 backups.
* Preserved the existing backup naming format and creation behavior.
* Preserved `OSError` handling during rotation.
* Added a regression test covering backward clock skew.
* Added documentation explaining why filesystem timestamps are used instead of relying on filename timestamps.

## Regression Test

Added `test_backup_rotation_clock_skew`.

The test creates backups in filesystem order while simulating a backward-moving wall clock for the timestamps embedded in their filenames. This creates an inversion where a newer backup has an earlier filename.

The test verifies that:

* rotation retains exactly 10 backups;
* the genuinely oldest backups are removed;
* the newer backup with the misleading filename is preserved;
* the behavior fails with the previous lexicographic ordering and passes with the filesystem-timestamp ordering.

## Validation

* `tests/test_backup.py`: 5/5 passed.
* Existing backup rotation tests pass.
* Backup rotation failure handling test passes.
* Additional runnable test modules passed.
* `git diff --check` passes.
* Ruff reports only two pre-existing unused imports unrelated to this change.

There are pre-existing Windows-only symlink test failures and environment-related full-suite collection issues on this machine; these were confirmed to be unrelated to this change.

## Platform Note

`os.path.getctime()` represents creation time on Windows but metadata change time on POSIX systems. The implementation follows Issue #429's explicit requirement to use `getctime()` (or an equivalent filesystem timestamp) rather than relying solely on the wall-clock timestamp embedded in the backup filename.

### DIFF
--- a/termstory/backup.py
+++ b/termstory/backup.py
@@ -22,8 +22,9 @@ def backup_db() -> str:
 
     Backup filenames use a microsecond-precision timestamp
     (``termstory_backup_YYYYMMDD_HHMMSS_mmmmmm.db``) so back-to-back backups
-    never collide. The fixed-width timestamp keeps filenames lexicographically
-    sortable, which backup rotation relies on to find the oldest backups.
+    never collide. Rotation finds the oldest backups using the filesystem
+    timestamp rather than the timestamp embedded in the filename, because the
+    wall clock can jump backward between backups.
 
     Returns:
         The absolute path to the created backup file.
@@ -49,7 +50,15 @@ def backup_db() -> str:
 
     # Rotate backups: keep at most 10 latest backups
     try:
-        backups = sorted(glob.glob(os.path.join(backup_dir, "termstory_backup_*.db")))
+        # Order by filesystem timestamp (getctime) rather than by the timestamp
+        # embedded in the filename. If the wall clock runs backward (NTP
+        # correction, VM snapshot restore), a *newer* backup can receive an
+        # earlier name; lexicographic ordering would then rotate away the newest
+        # backup instead of the oldest.
+        backups = sorted(
+            glob.glob(os.path.join(backup_dir, "termstory_backup_*.db")),
+            key=os.path.getctime,
+        )
         while len(backups) > 10:
             oldest = backups.pop(0)
             if os.path.exists(oldest):
--- a/termstory/backup.py
+++ b/termstory/backup.py
@@ -17,14 +17,28 @@ def _get_backup_dir() -> str:
     return backup_dir
 
 
+def _backup_creation_key(path: str) -> tuple:
+    """Return a stable sorting key expressing backup creation order.
+
+    Prefer the filesystem birth/creation timestamp where it is reliably exposed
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

deepmhatre13:
@bitflicker64 The Python 3.9 CI failure appears unrelated to PR #441, and I investigated it in detail.

* PR #441 has a zero-character diff for both `termstory/reminder.py` and `tests/test_reminder.py`.
* The failing test, `test_start_sleep_daemon_concurrent_invocations_spawn_once`, and its implementation already exist on `main` from the earlier sleep-daemon concurrency change (#424).
* The test passes on `main` and on this branch locally. The failure appears to be timing-sensitive and Linux/Python-3.9 scheduling-dependent.
* The traceback is consistent with a pre-existing TOCTOU race in `start_sleep_daemon()`: the PID file is created empty with `O_EXCL`, and the PID is written afterward. A concurrent invocation can observe the empty file as stale, remove it, and subsequently spawn a second daemon.
* I did not modify `reminder.py` or `tests/test_reminder.py`, since fixing that unrelated race would be outside the scope of #429.

The #429 changes remain unchanged and validated: all 5 backup tests pass, including the deterministic clock-skew regression test, and `git diff --check` is clean.

I recommend keeping this PR scoped to the backup-rotation fix. The reminder PID-file race can be addressed separately if needed.

bitflicker64:
## Review summary

**Label check:** ✅ `ECSOC26` and `area/storage` present.

**CI:** Python 3.9 failure is pre-existing — explained in detail by @deepmhatre13 on the PR (TOCTOU race from #424, zero-diff files in this PR). Not a blocker for this PR.

**Review:** CodeRabbit re-reviewed after fixes and switched to `APPROVED`. No outstanding actionable comments.

**Merge-ready** once CI re-runs green across all versions (the 3.9 failure is unrelated).

PR #442, opened 2026-08-20
fix: bound git status for large untracked trees

### LINKED ISSUE #431: fix: MCP snapshot capture_git_status hangs on repos with large untracked trees
## Bug Report

**Describe the bug**
`capture_git_status()` calls `git status --porcelain` without `--untracked-files=no`, causing it to hang or take minutes on repos with large untracked directory trees (e.g. `node_modules` not in `.gitignore`, build artifacts).

**Steps to reproduce**
1. Create a repo with a large untracked directory (100k+ files)
2. Trigger an MCP snapshot
3. Observe the snapshot taking 30+ seconds

**Expected behavior**
Add a timeout or use `--untracked-files=normal` (default) with a reasonable file count cap.

deepmhatre13:
## Summary

`capture_git_status()` could take excessive time — or effectively hang — on repositories with large untracked directory trees (e.g. `node_modules`, build artifacts), because it invoked `git status --porcelain` via a blocking `subprocess.run()` call that buffered the entire output before any bounding was applied. This PR bounds both Git's own traversal and the memory/CPU cost of processing its output, so MCP snapshot capture stays fast even on pathological repos.

## Changes

**Core (`termstory/mcp_snapshot.py`)**
- Added `--untracked-files=normal` to the `git status` invocation so huge untracked directories are collapsed to a single `?? dir/` entry instead of being recursively enumerated .
- Switched from a blocking `subprocess.run()` collect-then-cap approach to a streaming `subprocess.Popen()` read, so the retained list is capped at `_MAX_UNCOMMITTED_FILES` (1000) as lines arrive rather than after the full output is buffered in memory .
- When the cap is hit mid-stream, the Git process is terminated (with a fallback `kill()` on timeout) and `uncommitted_files_truncated` is set on the result .
- Preserved existing `is_repo`/`branch`/`uncommitted_files` structure and the `TimeoutExpired`/`OSError` exception handling .

**Tests (`tests/test_mcp_snapshot.py`)**
- Updated `_git_status_side_effect` and added `_git_status_popen_side_effect` to reflect the new `Popen`-based status call and assert `--untracked-files=normal` is present .
- Added regression coverage for the large-untracked-tree/truncation scenario, verifying the `Popen` command args, the 1000-entry cap, `uncommitted_files_truncated`, and that `terminate()` is called .

## Architecture / Flow

```mermaid
flowchart LR
  A["capture_git_status(cwd)"] --> B["Popen: git status --porcelain --untracked-files=normal"]
  B --> C{"line count >= 1000?"}
  C -- "no" --> D["append scrubbed line"]
  D --> B
  C -- "yes" --> E["terminate() / kill() on timeout"]
  E --> F["uncommitted_files_truncated = True"]
  D --> G["wait(timeout=5)"]
  F --> H["result"]
  G --> H
```

## Fixes

- Fixes #431 — `capture_git_status()` no longer performs unbounded work on large untracked directory trees.

## Testing

- `pytest tests/test_mcp_snapshot.py`
- Regression scenario simulates 1500 scattered untracked files plus a tracked modification and asserts the retained list is capped at 1000 with truncation flagged .

## Notes

- Review flagged a **duplicate test function name**: `test_capture_git_status_bounds_large_untracked_tree` is defined twice in `tests/test_mcp_snapshot.py`  — since it's the same module, the second definition silently shadows the first, so the `subprocess.run`-based version (exercising 1500 files via `CompletedProcess`) never actually runs. **One of these needs to be renamed before merge** to get the claimed coverage.
- Greptile raised a concern that truncated snapshots may still present the capped 1000 entries as if they were the complete uncommitted-file count downstream — worth confirming any consumer of `uncommitted_files_truncated` (e.g. formatting/display layer) actually surfaces the truncation to the user rather than silently showing a partial list as complete.

### DIFF
--- a/termstory/mcp_snapshot.py
+++ b/termstory/mcp_snapshot.py
@@ -11,6 +11,14 @@ from termstory.sanitizer import redact_command
 
 logger = logging.getLogger(__name__)
 
+# Cap the number of uncommitted-file entries retained in a single snapshot. A
+# `git status` listing can otherwise surface tens of thousands of untracked
+# files (e.g. scattered build artifacts), producing an enormous snapshot and
+# heavy per-entry scrubbing. `--untracked-files=normal` already collapses each
+# huge *untracked directory* (node_modules/, dist/, build/) to a single entry,
+# so this cap only bounds the pathological scattered-file case.
+_MAX_UNCOMMITTED_FILES = 1000
+
 # IDE/editor environment variable prefixes. Their *presence* drives IDE
 # detection, but their values are never persisted: they routinely hold socket
 # paths (e.g. VSCODE_IPC_HOOK_CLI, NVIM_LISTEN_ADDRESS), askpass helpers,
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

bitflicker64:
## Review summary

**Label check:** ✅ `ECSOC26` and `area/storage` present.

**CI:** All 4 Python versions are failing on this PR. Not mergeable until the test suite passes.

**Review blockers (2):**

1. **CodeRabbit** — `capture_git_status()` collects the full `git status` output into memory via `subprocess.run()` before applying the 1000-entry cap. A repo with many scattered untracked files can still allocate an unbounded buffer. Needs a streaming collection strategy that stops the Git process after the cap is reached.

2. **Greptile** — When Git returns more than 1000 entries, the snapshot silently discards the remainder without marking the snapshot as truncated. The formatter then presents 1000 as the total count. Needs a `truncated` flag so consumers know the data is incomplete.

Both are valid. This PR needs fixes before it can merge.

deepmhatre13:
:miku /ecsoc

bitflicker64 (approved):
LGTM. Bounded streaming Popen implementation and truncation indicator address all previous review blockers, and all CI tests are passing.

PR #443, opened 2026-08-23
feat: add stats json output

### LINKED ISSUE #430: feat: add termstory stats --json for machine-readable session statistics
## Feature Request

**Is your feature request related to a problem?**
There's no machine-readable way to get session statistics. `termstory stats` only outputs human-readable text.

**Describe the solution you'd like**
Add `termstory stats --json` that outputs structured JSON with total sessions, commands, projects, time ranges, and per-project breakdowns.

**Use case**
CI/CD dashboards, personal productivity tracking scripts, integration with other tools.

deepmhatre13:
## Summary

- Add `termstory stats --json` for machine-readable session statistics.
- Add reusable raw statistics assembly in `termstory/stats.py`.
- Preserve the existing human-readable `termstory stats` output.
- Add populated and empty-database unit and CLI coverage.
- Document the new `--json` option.

## Validation

- `python -m pytest tests/test_stats.py` — 20 passed
- `python -m pytest tests/test_stats.py tests/test_database_queries.py tests/test_save_data_none_safety.py` — 26 passed
- `python -m pytest tests/test_exporter.py tests/test_insights.py` — 57 passed
- `python -m pytest tests/test_notebook.py tests/test_timeline.py` — 15 passed
- Ruff passed for the new stats implementation.
- `git diff --check` passed.
- CliRunner end-to-end validation passed for populated and empty databases.
- JSON stdout was verified to be parseable and free of ANSI/Rich output.
- Existing human-readable stats behavior was preserved.

There are pre-existing Windows-only test collection issues involving `os.geteuid()` and `test_bash_inspect.py`; these are unrelated to this change.

Closes #430

### DIFF
--- a/termstory/stats.py
+++ b/termstory/stats.py
@@ -157,6 +157,92 @@ def project_breakdown(db: Database) -> Dict[str, Dict[str, Any]]:
         
     return breakdown
 
+def stats_json(db: Database) -> Dict[str, Any]:
+    """Assemble a plain, machine-readable statistics payload.
+
+    Reuses the existing ``project_breakdown()`` calculation (including the
+    ``General / No Project`` -> ``Other`` mapping) instead of duplicating it, and
+    computes the top-level session/command/project totals plus the overall activity
+    time window directly from the database.
+
+    All return values are plain JSON-compatible primitives (ints, strings or None) so
+    the result can be passed straight to ``json.dumps()`` with no custom serializer.
+
+    Returns a dict shaped as::
+
+        {
+            "total_sessions": int,
+            "total_commands": int,
+            "total_projects": int,
+            "time_range": {"earliest": str | None, "latest": str | None},
+            "projects": [
+                {"name", "id", "path", "commands_count", "total_duration",
+                 "sessions_count", "first_seen", "last_seen"}, ...
+            ],
+        }
+
+    For an empty database, all totals are 0, ``projects`` is ``[]`` and the time
+    range is ``{"earliest": None, "latest": None}``.
+    """
+    conn = db.get_connection()
+    try:
+        cursor = conn.cursor()
+        cursor.execute("SELECT COUNT(*) FROM sessions")
+        total_sessions = cursor.fetchone()[0]
+        cursor.execute("SELECT COUNT(*) FROM commands")
+        total_commands = cursor.fetchone()[0]
+        cursor.execute("SELECT COUNT(*) FROM projects")
+        total_projects = cursor.fetchone()[0]
+
+        # Overall activity window covering both sessions and commands.
+        cursor.execute(
+            "SELECT MIN(start_time), MAX(end_time) FROM sessions "
+            "WHERE start_time IS NOT NULL"
+        )
+        s_min, s_max = cursor.fetchone()
+        cursor.execute(
+            "SELECT MIN(timestamp), MAX(timestamp) FROM commands "
+            "WHERE timestamp IS NOT NULL"
+        )
+        c_min, c_max = cursor.fetchone()
+    finally:
+        conn.close()
+
+    times = [t for t in (s_min, s_max, c_min, c_max) if t is not None]
+    if times:
+        earliest = datetime.fromtimestamp(min(times)).isoformat()
+        latest = datetime.fromtimestamp(max(times)).isoformat()
+    else:
+        earliest = None
+        latest = None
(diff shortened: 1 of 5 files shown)


deepmhatre13:
:miku /ecsoc

termstory/stats.py
```diff
@@ -157,6 +157,216 @@ def project_breakdown(db: Database) -> Dict[str, Dict[str, Any]]:
...
+        )
+        session_stats = {row[0]: (row[1], row[2] or 0) for row in cursor.fetchall()}
+
+        # First/last activity for project-less rows (mirrors project_breakdown).
+        cursor.execute(
+            "SELECT MIN(timestamp), MAX(timestamp) FROM commands WHERE project_id IS NULL"
+        )
+        c_min, c_max = cursor.fetchone()
+        cursor.execute(
+            "SELECT MIN(start_time), MAX(end_time) FROM sessions WHERE project_id IS NULL"
```
bitflicker64, line 226: Small leftover of the bug fixed above: the `Other` bucket's `last_seen` still comes from `MAX(end_time)`, so an unfinished project-less session does not bound it, and a single payload ends up internally inconsistent:

```
sessions (project_id NULL): (1000, 2000), (9000, NULL)
time_range.latest         -> ts 9000
projects[Other].last_seen -> ts 2000
```

Named projects are unaffected, since their `first_seen`/`last_seen` come from the `projects` row.

Worth noting `project_breakdown()` has the identical query at line 136, so changing only this one breaks the mirroring between the two functions. Either fix both or leave both for a follow-up; I would not hold this PR for it. If you do take it here:

```suggestion
            "SELECT MIN(start_time), MAX(COALESCE(end_time, start_time)) "
            "FROM sessions WHERE project_id IS NULL"
```

termstory/stats.py
```diff
@@ -157,6 +157,216 @@ def project_breakdown(db: Database) -> Dict[str, Dict[str, Any]]:
...
+
+    Returns ``None`` for ``None``/out-of-range input instead of raising, following
+    the existing repository convention (see ``insights.analyze_all``, which wraps
+    ``datetime.fromtimestamp`` in ``except (OSError, ValueError, OverflowError)``)
+    so a single malformed row cannot crash ``stats --json``.
+    """
+    if ts is None:
+        return None
+    try:
+        return datetime.fromtimestamp(ts)
```
bitflicker64, line 181: Two small things about this conversion, both follow-up material.

Plain `datetime.fromtimestamp(ts)` raises `OSError` on Windows for negative and very large values where `exporter._safe_fromtimestamp()` succeeds (that helper converts through `tz=timezone.utc` then back to naive local specifically as a Windows workaround). So on Windows `stats --json` will emit `null` for some timestamps that `termstory export` renders fine. Reusing that helper would make the two paths agree.

Separately, the result is naive local time with no UTC offset, so `earliest`, `latest`, `first_seen`, and `last_seen` denote different instants depending on the machine that ran the command, while `docs/cli-reference.md` calls them ISO-8601. Since this is a documented contract for dashboards and CI, one docs line saying the timestamps are naive local time would save a consumer the surprise.

bitflicker64 (approved):
The follow-up commit addresses everything blocking.

**`time_range.latest` and unfinished sessions**: fixed. The `UNION ALL` over session starts, non-NULL session ends, and command timestamps means a still-running session bounds `latest` via its `start_time`. The earlier repro (sessions `(1000, 2000)` and `(9000, NULL)`) now yields ts 9000. Doing the min/max in Python rather than in SQL is the right trade, since it lets one out-of-range row be skipped instead of masking every valid timestamp.

**Same-name projects**: fixed. `project_breakdown_json()` keys by `projects.id`, so two projects named `demo` now emit two entries with correct `id`/`path` pairs, and `len(projects)` matches `total_projects`. It runs the same underlying queries as `project_breakdown()` (same `cmd_counts`, same `session_stats`, same `first_seen`/`last_seen` source), so any difference between the two views comes from grouping rather than from a second, drifting calculation.

**Crash resilience**: `_safe_datetime_from_ts` follows the `insights.analyze_all` precedent, and the new tests cover the unfinished-session, duplicate-name, and out-of-range cases directly. `tests/test_stats.py` and `tests/test_cli_commands.py` are 60 passed locally, and CI is green on 3.9 through 3.12.

Two non-blocking notes inline. Merging.

PR #455, opened 2026-08-26
fix(git_integration): validate git worktree status

### LINKED ISSUE #445: fix(git_integration): is_git_repo returns True inside a .git directory and for bare repos
## Context

`is_git_repo()` in `termstory/git_integration.py:5` gates every git lookup in `get_project_commits()` and `get_timeframe_git_stats()`. Its docstring says it checks for "a valid git repository worktree".

## Problem

It decides using only the exit code:

```python
res = subprocess.run(["git", "-C", abs_path, "rev-parse", "--is-inside-work-tree"], ...)
return res.returncode == 0
```

`git rev-parse --is-inside-work-tree` exits 0 and prints `false` when the path is inside a `.git` directory or inside a bare repository. The stdout value is the actual answer and it is discarded.

Reproduction from the repo root:

```
$ git -C .git rev-parse --is-inside-work-tree
false
$ python -c "from termstory.git_integration import is_git_repo; print(is_git_repo('.git'))"
True
```

So a project path pointing at a bare repo or a `.git` directory is treated as a worktree, and the subsequent `git log` calls return nothing useful while still paying the subprocess cost.

## Expected

Return True only when the command exits 0 **and** stdout is `true`.

## Acceptance criteria

- [ ] `is_git_repo` checks `res.stdout.strip() == "true"` alongside the return code
- [ ] `is_git_repo("<repo>/.git")` returns False
- [ ] Existing tests at `tests/test_git_integration.py:34` and `:47` still pass
- [ ] A regression test for the `.git` directory case added to `tests/test_git_integration.py`

## Scope

One function, one added condition, plus a test. Roughly 1 file, 8 lines.

deepmhatre13:
Closes #445
## Summary

Fixes #445 by making `is_git_repo()` verify the actual result reported by Git instead of relying only on the subprocess exit code.

Previously, `git rev-parse --is-inside-work-tree` could exit successfully with return code `0` while printing `false` when the supplied path was a `.git` directory or a bare repository. Since `is_git_repo()` only checked the return code, these paths were incorrectly treated as valid Git worktrees.

### Changes

- Updated `is_git_repo()` to require both:
  - `res.returncode == 0`
  - `res.stdout.strip() == "true"`
- Added a regression test verifying that a normal repository root returns `True`.
- Added a regression test verifying that the repository's `.git` directory returns `False`.
- Added mocked positive and negative stdout cases to verify both sides of the new condition.
- Kept the change limited to `termstory/git_integration.py` and `tests/test_git_integration.py`.
- Did not modify the analogous logic in `mcp_snapshot.py`, as it is outside the scope of this issue.

### Testing

Targeted Git integration tests:

    4 passed in 1.75s

`git diff --check`:

    clean

The full test suite was also attempted. Collection is currently affected by pre-existing environment/platform issues, including a missing `textual` dependency, use of POSIX-only `os.geteuid` on Windows, and Windows file-lock errors. These failures are unrelated to this change.

### Scope

This is a focused bugfix with no API changes, migrations, or unrelated refactoring.

### DIFF
--- a/termstory/git_integration.py
+++ b/termstory/git_integration.py
@@ -19,7 +19,10 @@ def is_git_repo(path: str) -> bool:
             check=False,
             timeout=10
         )
-        return res.returncode == 0
+        # Returning True only when both the process succeeded and git reports
+        # that we are inside a work tree. `rev-parse --is-inside-work-tree`
+        # exits 0 and prints "false" for a .git directory or a bare repo.
+        return res.returncode == 0 and res.stdout.strip() == "true"
     except Exception:
         return False
 
--- a/tests/test_git_integration.py
+++ b/tests/test_git_integration.py
@@ -64,6 +64,22 @@ def test_git_operations_on_temp_repo(tmp_path):
     assert len(commits[0]["hash"]) == 40
     assert commits[0]["timestamp"] > 0
 
+def test_is_git_repo_worktree_vs_git_dir(tmp_path):
+    # Initialize a temporary git repository
+    repo_path = tmp_path / "repo"
+    repo_path.mkdir()
+    try:
+        subprocess.run(["git", "init"], cwd=str(repo_path), check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    except Exception:
+        # If git is not installed or init fails, skip the rest of the test
+        return
+
+    # The repository root is a valid worktree
+    assert is_git_repo(str(repo_path)) is True
+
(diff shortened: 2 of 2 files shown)


bitflicker64 (approved):
Solid fix. Checking stdout.strip() == "true" in addition to returncode handles the .git dir and bare-repo cases that previously slipped through. Regression tests in test_git_integration.py cover both the real-repo and mocked false/true paths. Does what #445 asked.

deepmhatre13:
:miku /ecsoc

PR #456, opened 2026-08-26
feat(tui): scope heatmap highlights to visible window

### LINKED ISSUE #446: feat(tui): scope heatmap personal-best to the visible window in All History mode
## Context

`_compute_highlight_days()` in `termstory/tui.py:135` decides which days get the pulse animation in the heatmap. A day qualifies if its command count ties the personal best across the visible window, or if it contains a session of 8+ hours.

## Problem

This is written up as a known limitation in the function's own docstring at `termstory/tui.py:150`:

> When `days_limit` is `None` ("All History" mode), the personal-best computation scans the ENTIRE session history rather than a bounded window. This is correct but may produce surprising results for long-time users (a single prolific day months ago suppresses all other days).

In All History mode the highlight collapses onto one historical day and nothing in the recent window ever lights up, so the animation stops carrying information for exactly the users with the most history.

## Expected

Scope the personal-best calculation to the days currently rendered. The docstring notes the blocker: `days_limit` is not currently threaded through to `_compute_highlight_days`, so the call chain from `generate_heatmap()` at `termstory/tui.py:170` needs to pass it down.

## Acceptance criteria

- [ ] `_compute_highlight_days` accepts the visible window and computes `max_count` over only those days
- [ ] `days_limit` is threaded from `generate_heatmap` to the call site
- [ ] With `days_limit=None`, a prolific day outside the rendered window no longer suppresses highlights inside it
- [ ] The 8+ hour session rule is unchanged
- [ ] Test added to `tests/test_tui.py`
- [ ] The TODO block in the docstring is removed

## Scope

One function signature plus its call chain. Roughly 1 file, 30 lines.

deepmhatre13:
Closes #446

## Summary

This PR fixes the heatmap personal-best highlighting behavior so that the personal-best command count is calculated only from the days currently visible in the rendered heatmap window.

Previously, `_compute_highlight_days()` calculated the personal-best using all available `day_counts`. In All History mode, all historical sessions are loaded, so a single highly active day from months earlier could set the personal-best threshold and prevent recent visible days from being highlighted.

The fix scopes the personal-best calculation to the same resolved window used by the heatmap renderer.

## Changes

- Updated `_compute_highlight_days()` to accept the resolved visible `days_limit`.
- Constructed the visible date window using the same current-date source and day-range semantics as `generate_heatmap()`.
- Filtered `day_counts` to visible dates before calculating `max_count`.
- Preserved the existing session data; historical sessions are not removed or filtered globally.
- Updated both production call paths:
  - `calculate_dashboard_stats()`
  - `TermStoryWorkspace.update_stats_header()` highlight caching
- Updated existing `_compute_highlight_days()` test calls for the new internal signature.
- Added regression coverage for a prolific historical day being outside the visible window.
- Added end-to-end coverage for the actual `days_limit=None` / All History workspace path.
- Removed the obsolete TODO describing the previous limitation.
- Added a comment documenting that the highlight calculation must remain synchronized with the rendered heatmap window.

## All History behavior

`days_limit=None` represents All History for the data scope, so all historical sessions are loaded.

The existing dashboard rendering path resolves this to the existing 90-day rendered window before generating the heatmap. The highlight calculation now uses that same resolved window, ensuring that personal-best highlights correspond to the days actually displayed.

Therefore:

- Historical data remains available to the application.
- The heatmap continues to render its existing 90-day strip.
- A prolific day outside that rendered strip cannot determine the personal-best threshold.
- Recent visible days can correctly receive personal-best highlights.

## 8+ Hour Session Rule

The existing `EIGHT_HOURS_SECONDS` highlighting rule is unchanged.

Sessions lasting at least 8 hours continue to be highlighted independently of the personal-best command-count calculation.

### Focused tests

`tests/test_tui_cyberpunk_animations.py`

- 30 passed

Relevant `tests/test_tui.py` tests:

- 5 passed

Coverage includes:

- Personal-best highlighting within a visible window
- Tied personal-best days
- Historical prolific days outside the visible window
- All History workspace behavior
- Highlight cache behavior
- 8-hour session highlighting
- Existing heatmap behavior

`git diff --check` also passes with no whitespace errors.

## Scope

Changes are limited to:

- `termstory/tui.py`
- `tests/test_tui.py`
- `tests/test_tui_cyberpunk_animations.py`

No unrelated modules or systems were modified.

### DIFF
--- a/termstory/tui.py
+++ b/termstory/tui.py
@@ -133,7 +133,11 @@ def calculate_streak(sessions: List[Session]) -> int:
 EIGHT_HOURS_SECONDS = 8 * 3600
 
 
-def _compute_highlight_days(day_counts: dict, sessions: List[Session]) -> set:
+def _compute_highlight_days(
+    day_counts: dict,
+    sessions: List[Session],
+    days_limit: int = 30,
+) -> set:
     """Compute the set of dates that should be eligible for the magenta→pink
     pulse animation (issue #42).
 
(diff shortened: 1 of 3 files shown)


deepmhatre13:
:miku /ecsoc

deepmhatre13:
@bitflicker64 kindly review the pr

termstory/tui.py
```diff
@@ -144,22 +148,39 @@ def _compute_highlight_days(day_counts: dict, sessions: List[Session]) -> set:
...
+    own).
     """
-    if not day_counts:
+    if not day_counts or days_limit is None or days_limit <= 0:
         return set()
 
-    max_count = max(day_counts.values())
-    highlight = {d for d, c in day_counts.items() if c > 0 and c == max_count}
+    now = get_current_time().date()
+    visible_days = {now - timedelta(days=i) for i in range(days_limit)}
```
bitflicker64, line 171: Personal-best is now computed from visible_day_counts only, using the same today-(days_limit-1)..today window generate_heatmap renders. That closes the All History suppression case in #446.

termstory/tui.py
```diff
@@ -3121,21 +3146,38 @@ def update_stats_header(self) -> None:
         # This avoids re-iterating all sessions every 0.5s — important for not
         # perturbing the timing of background worker threads (e.g. the exec
         # review worker) on slower Python versions (3.9/3.10).
+        #
+        # Issue #446: highlights are now scoped to the VISIBLE heatmap window,
+        # so the cache is also keyed on the current date (midnight rolls the
+        # rendered window forward even with no new sessions) and on the
+        # resolved window (--days/--all-history flips it).
+        current_date = get_current_time().date()
+        resolved_window = self.days_limit or 90
```
bitflicker64, line 3155: Caching on current_date and resolved_window is required once highlights are window-scoped; otherwise midnight or --days changes would keep a stale set. Looks correct.

bitflicker64 (approved):
Personal-best highlighting is scoped to the rendered window, including the All History path that resolves None to 90 days. Session data is still loaded in full. Cache keys now include date and window. Matches #446. CI is green.

PR #459, opened 2026-08-26
fix(search): preserve per-command project attribution

### LINKED ISSUE #457: Title: fix(search): preserve per-command project attribution in search results
## Context

TermStory supports per-command project attribution so that a session which changes directories between projects can retain the correct project context for each command.

The repository architecture explicitly treats per-command project attribution as important for search, insights, and exports. Search also supports project-scoped queries such as:

termstory search stripe --project "Acme Billing"

## Problem

Search result construction can lose the per-command project context when results are aggregated back into session-level records.

This is problematic for sessions that contain commands from multiple projects:

1. A session starts in Project A.
2. The user changes into Project B.
3. Commands from both projects are recorded in the same session.
4. Search matches a command belonging specifically to Project B.
5. The result is reconstructed using the session-level project rather than the command-level attribution.

The resulting search output can therefore associate a matching command with the wrong project or make project-scoped filtering inconsistent with the underlying command attribution.

## Expected

Search results should preserve the most specific project attribution available for the matched command.

For a command with explicit per-command project attribution:

- search should report/filter against that project;
- the session's fallback project should not override it;
- commands from different projects within the same session must remain distinguishable.

If a command has no per-command project attribution, the existing session-level fallback behavior should remain unchanged.

## Acceptance criteria

- Search results preserve per-command project attribution when it exists.
- A session containing commands from two different projects can return results associated with the correct project.
- `--project` filtering uses the command's project attribution where available.
- Existing session-level project behavior remains unchanged for commands without explicit attribution.
- Regression tests cover a single session containing commands attributed to two different projects.
- Existing search tests continue to pass.

## Scope

Keep the change limited to the search/result construction path and its tests.

Do not redesign project detection or database schema.

If this aligns with the project’s needs and scope, I’d be happy to work on it. Please assign it to me.

deepmhatre13:
Closes #457
## Summary

Fix search result construction so that when a matched command has explicit per-command project attribution, the search result reports that project instead of incorrectly using the session's final project.

Previously, the search queries could use per-command projects for filtering, but the returned result was built from the session-level `project_id` and project metadata. This caused commands from different projects within the same session to be attributed to the wrong project.

### Changes

- Updated all three search paths in `termstory/search.py`:
  - `_search_new_fts5`
  - `_search_fts5`
  - `_search_standard`
- Preserve the project associated with the matched command when explicit `project_id` attribution exists.
- Fall back to the session-level project when the matched command has no project attribution.
- Keep the existing one-result-per-session behavior.
- Added regression coverage for:
  - A single session containing commands from two different projects.
  - Correct project attribution for different search matches.
  - `--project` filtering combined with correct result attribution.
  - Session-level project fallback when command-level attribution is absent.
  - All three search backends.

### Validation

- `tests/test_search.py` — **5 passed**
- Related FTS5/ASK/formatter/RAG tests — **71 passed**
- Integration/database tests — **4 passed**
- `git diff --check` — clean
- No database schema or project-detection changes.

### DIFF
--- a/termstory/search.py
+++ b/termstory/search.py
@@ -99,15 +99,31 @@ def _search_new_fts5(
             FROM matched_session_ids
             GROUP BY id
         )
-        SELECT s.id, s.start_time, s.end_time, s.duration_seconds, s.project_id, p.name, p.path, s.ai_summary
+        SELECT s.id, s.start_time, s.end_time, s.duration_seconds,
+               COALESCE(ep.id, s.project_id),
+               COALESCE(ep.name, p.name),
+               COALESCE(ep.path, p.path),
+               s.ai_summary
         FROM sessions s
         JOIN best_matches bm ON s.id = bm.id
         LEFT JOIN projects p ON s.project_id = p.id
+        LEFT JOIN projects ep ON ep.id = (
+            -- #457: attribute the result to the project of the earliest
+            -- command that actually matched the query and carries explicit
+            -- per-command attribution; NULL -> fall back to session project.
+            SELECT c.project_id
+            FROM commands c
+            WHERE c.session_id = s.id
+              AND c.id IN (SELECT rowid FROM commands_fts WHERE commands_fts MATCH ?)
+              AND c.project_id IS NOT NULL
+            ORDER BY c.timestamp ASC, c.id ASC
+            LIMIT 1
+        )
         LEFT JOIN commands cmd_per_proj ON cmd_per_proj.session_id = s.id
         LEFT JOIN projects p2 ON cmd_per_proj.project_id = p2.id
         WHERE 1=1
     """
-    params = [fts_query, fts_query, fts_query]
+    params = [fts_query, fts_query, fts_query, fts_query]
 
     if project_filter:
         # Match if the session's final project OR any per-command project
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

termstory/search.py
```diff
@@ -99,15 +99,31 @@ def _search_new_fts5(
...
+        SELECT s.id, s.start_time, s.end_time, s.duration_seconds,
+               COALESCE(ep.id, s.project_id),
+               COALESCE(ep.name, p.name),
+               COALESCE(ep.path, p.path),
+               s.ai_summary
         FROM sessions s
         JOIN best_matches bm ON s.id = bm.id
         LEFT JOIN projects p ON s.project_id = p.id
+        LEFT JOIN projects ep ON ep.id = (
+            -- #457: attribute the result to the project of the earliest
```
bitflicker64, line 111: This subquery picks the earliest matching command that has an explicit project_id, then COALESCE falls back to the session project. That is exactly what #457 asked for on the new FTS5 path.

tests/test_search.py
```diff
@@ -145,6 +145,144 @@ def test_project_filter_matches_per_command_project(tmp_path, monkeypatch):
...
+    from termstory.search import advanced_search
+
+    def run_all_backends(query, **kwargs):
+        """Yield (backend_name, results) for the two FTS search implementations."""
+        # _search_new_fts5
+        yield "new_fts5", advanced_search(db, query=query, fts=True, **kwargs)
+        # _search_fts5
+        yield "fts5", advanced_search(db, query=query, **kwargs)
+
+    for backend, results in run_all_backends("stripe"):
```
bitflicker64, line 192: The mixed-session fixture (stripe in Acme, npm in Mobile, session.project_id = Mobile) is the right regression. Covering new_fts5, fts5, and the dropped-index standard path is enough.

bitflicker64 (approved):
All three search backends now report the matched command's project when that command has an explicit project_id, and fall back to the session project otherwise. One-result-per-session is unchanged. Tests cover the mixed-project session from #457. CI is green.

PR #470, opened 2026-08-26
fix(tags): prevent network utilities from matching debug

### LINKED ISSUE #452: fix(tags): debug tag matches curl, ping, dig, and nslookup as debug work
## Context

`termstory/tags.py` TAG_RULES for "debug" includes cmd_patterns such as `\bcurl\b`, `\bping\b`, `\bdig\b`, `\bnslookup\b` (see the debug block near the top of the file). Those patterns are compiled case-insensitively and applied to every command string.

## Problem

Everyday network checks (`curl https://example.com`, `ping 8.8.8.8`) are tagged as debug. That pollutes session tags, skews any UI or search that filters by tag, and makes the debug tag less useful for actual debugger/profiler work.

## Expected

Debug cmd patterns should target real debug tooling (pdb, gdb, lldb, strace, valgrind, docker logs, journalctl, etc.) and not bare network utilities. Curl/ping can stay only when paired with stronger signals if needed, or move to a separate ops/network tag.

## Acceptance criteria

- [ ] `curl`, `ping`, `dig`, and `nslookup` alone no longer produce the debug tag
- [ ] pdb/gdb/strace-style commands still produce debug
- [ ] Unit tests in `tests/test_tags.py` cover both the false-positive and true-positive cases

## Scope

1-2 files; roughly 15-40 lines.

deepmhatre13:
Closes #452
## Summary

The `debug` tag was incorrectly matching common network utilities such as `curl`, `ping`, `dig`, and `nslookup`.

## Changes

- Removed the four overly broad network-utility patterns from the `debug` command rules.
- Preserved the existing debugger/profiler patterns including `pdb`, `gdb`, `lldb`, `strace`, `valgrind`, and related tools.
- Added regression coverage ensuring:
  - `curl`, `ping`, `dig`, and `nslookup` do not produce the `debug` tag.
  - `pdb`, `gdb`, `lldb`, and `strace` continue to produce the `debug` tag.
- No changes to project detection, database schema, CLI/TUI behavior, or unrelated tag rules.

## Validation

- `pytest tests/test_tags.py -v` — 4 passed
- `git diff --check` — clean
- Final diff limited to:
  - `termstory/tags.py`
  - `tests/test_tags.py`

### DIFF
--- a/termstory/tags.py
+++ b/termstory/tags.py
@@ -17,7 +17,7 @@ TAG_RULES = {
         "cmd_patterns": [
             r"\bdebug(ger)?\b", r"\bpdb\b", r"\bipdb\b", r"\blldb\b", r"\bgdb\b",
             r"\btrace\b", r"\bvalgrind\b", r"\bstrace\b", r"\bprofile(r)?\b",
-            r"\bpyinstrument\b", r"\bcurl\b", r"\bping\b", r"\bdig\b", r"\bnslookup\b",
+            r"\bpyinstrument\b",
             r"\btcpdump\b", r"\bdoctor\b", r"\bdiagnose\b", r"\bdiagnostic\b",
             r"\bjournalctl\b", r"\bdmesg\b", r"\bdocker logs\b"
         ],
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -32,6 +32,33 @@ def test_compute_tags_from_text():
     # Ordering should be: deploy, debug, setup, test, docs
     assert tags == ["deploy", "setup", "test"]
 
+def test_network_utilities_not_tagged_debug():
+    """Regression test for issue #452: ordinary network utilities must not
+    receive the debug tag. Assertions check 'debug' specifically so another
+    tag cannot mask the result."""
+    for cmd in [
+        "curl https://example.com",
+        "ping 8.8.8.8",
+        "dig example.com",
+        "nslookup example.com",
+    ]:
+        assert "debug" not in compute_tags_from_text([cmd], []), (
+            f"{cmd!r} must not be tagged debug"
+        )
+
+def test_debug_tools_still_tagged_debug():
+    """Regression test for issue #452: removing the bare network-utility
+    patterns must not affect legitimate debugger/profiler commands."""
+    for cmd in [
+        "python3 -m pdb script.py",
+        "gdb ./program",
+        "lldb ./program",
+        "strace ./program",
+    ]:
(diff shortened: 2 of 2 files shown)


deepmhatre13:
:miku /ecsoc

termstory/tags.py
```diff
@@ -17,7 +17,7 @@
         "cmd_patterns": [
             r"\bdebug(ger)?\b", r"\bpdb\b", r"\bipdb\b", r"\blldb\b", r"\bgdb\b",
             r"\btrace\b", r"\bvalgrind\b", r"\bstrace\b", r"\bprofile(r)?\b",
-            r"\bpyinstrument\b", r"\bcurl\b", r"\bping\b", r"\bdig\b", r"\bnslookup\b",
+            r"\bpyinstrument\b",
```
bitflicker64, line 20: Removing `curl`, `ping`, `dig`, and `nslookup` from the debug cmd_patterns is the right fix for #452. pdb/gdb/lldb/strace/valgrind stay in place, which matches the expected behavior.

tests/test_tags.py
```diff
@@ -32,6 +32,33 @@ def test_compute_tags_from_text():
     # Ordering should be: deploy, debug, setup, test, docs
     assert tags == ["deploy", "setup", "test"]
 
+def test_network_utilities_not_tagged_debug():
```
bitflicker64, line 35: Good coverage of the four false positives from #452. Checking that `debug` is absent (rather than comparing the full tag list) keeps the assertion focused.

tests/test_tags.py
```diff
@@ -32,6 +32,33 @@ def test_compute_tags_from_text():
...
+        "ping 8.8.8.8",
+        "dig example.com",
+        "nslookup example.com",
+    ]:
+        assert "debug" not in compute_tags_from_text([cmd], []), (
+            f"{cmd!r} must not be tagged debug"
+        )
+
+def test_debug_tools_still_tagged_debug():
+    """Regression test for issue #452: removing the bare network-utility
```
bitflicker64, line 50: True-positive cases for pdb, gdb, lldb, and strace look correct. That is the other half of the acceptance criteria.

bitflicker64 (approved):
Looks good. The diff does what #452 asked: drop curl/ping/dig/nslookup from debug cmd_patterns in termstory/tags.py, keep the real debugger/profiler patterns, and add both false-positive and true-positive coverage in tests/test_tags.py. Scope is the two files named in the issue. No unrelated changes. Lint and the Python 3.9-3.12 test jobs are green.

PR #471, opened 2026-08-26
fix(search): recover from stale FTS indexes

### LINKED ISSUE #467: fix(search): recover from stale or incomplete FTS indexes
## Context

Termstory maintains searchable command/session data through its FTS-backed search paths as well as the standard SQL search fallback. The search layer therefore has two separate representations of command data: the authoritative records in the database and the derived search index.

## Problem

The search implementation can encounter a state where the FTS index is stale or incomplete relative to the underlying command data.

Examples include:

- a command exists in the authoritative `commands` table but has no corresponding FTS entry;
- an FTS entry references data that has subsequently changed or been removed;
- an interrupted indexing/rebuild operation leaves only part of the searchable data indexed;
- a database restored or migrated from an older state has valid command records but an incomplete search index.

In these cases, FTS-backed searches can silently return fewer results than the authoritative data contains, while the standard SQL path can still find the missing records.

This is worse than a hard failure because users receive apparently valid but incomplete search results.

## Expected

Search should remain semantically correct when the FTS index is stale or incomplete.

The implementation should:

- distinguish between an unavailable FTS index and an incomplete/stale index;
- preserve the existing fast FTS path when the index is healthy;
- provide a deterministic recovery/fallback path when FTS cannot be trusted;
- avoid returning silently incomplete results;
- avoid unnecessarily rebuilding the entire index for every normal search;
- preserve existing search filters such as project, tag, date, and query matching.

The authoritative database records must remain the source of truth.

## Acceptance criteria

- A command inserted into the authoritative database but missing from the FTS index can still be found by search.
- An incomplete/stale index does not silently cause search results to disappear.
- A healthy FTS index continues to use the existing optimized path.
- Existing standard-search fallback behavior remains intact.
- Project/tag/date filters continue to behave consistently between recovered and normal searches.
- Regression tests explicitly construct an incomplete/stale index state and verify that the expected command is still returned.
- Tests cover both a healthy index and the recovery/fallback path.
- No database schema migration is introduced.

## Scope

Keep the change limited to the search/index consistency and recovery path.

Do not redesign the search CLI or result format, and do not change project attribution semantics.

- `termstory/search.py`
- FTS/index maintenance code
- relevant database/query helpers
- `tests/test_search.py`
- existing FTS tests

Before implementing the fix, trace where the FTS index is populated, updated, rebuilt, and queried so the recovery behavior is based on the actual lifecycle rather than adding a blind fallback.

deepmhatre13:
Closes #467
## Summary

Fix search behavior when the FTS index exists but is stale or incomplete relative to the authoritative database records.

Previously, the search layer only checked whether the FTS tables existed. A partially populated or stale index could therefore cause valid commands to disappear from search results without any error.

This change:

- Adds lightweight consistency checks for the relevant FTS backends.
- Detects incomplete `search_index` coverage against authoritative command records.
- Detects stale `commands_fts` coverage using its document shadow table.
- Falls back to the existing authoritative SQL search path when an FTS backend is inconsistent.
- Preserves the existing optimized FTS path when the index is healthy.
- Preserves the existing `OperationalError` fallback behavior.
- Keeps project, tag, date, query, and per-command project attribution semantics unchanged.
- Does not rebuild the FTS index during normal searches.
- Introduces no schema or migration changes.

## Tests

Added regression coverage for:

- healthy FTS searches;
- healthy searches with zero matches;
- commands missing from `search_index`;
- stale/incomplete `commands_fts`;
- project/date/tag filtering during recovery;
- existing `OperationalError` fallback behavior.

Validation:

- `pytest tests/test_search.py -v` — 9 passed
- `pytest tests/test_fts5.py tests/test_archive.py -v` — 6 passed
- Combined search/FTS validation — 15 passed
- `git diff --check` — clean

The Windows-only `tests/test_cli_commands.py` `os.geteuid` collection failure is pre-existing and unrelated.

### DIFF
--- a/termstory/search.py
+++ b/termstory/search.py
@@ -23,10 +23,20 @@ def advanced_search(
         cursor = conn.cursor()
         
         if fts and query:
-            try:
-                return _search_new_fts5(conn, query, project_filter, since_ts, until_ts, tag_filters, limit)
-            except sqlite3.OperationalError:
-                logger.warning("FTS5 advanced search failed, falling back", exc_info=True)
+            # #467: an FTS index can exist yet be stale/incomplete relative to
+            # the authoritative commands table (lost triggers, interrupted
+            # migration, restored backup). Searching such an index silently
+            # omits records, so verify coverage first and route around it.
+            if _commands_fts_is_consistent(conn):
+                try:
+                    return _search_new_fts5(conn, query, project_filter, since_ts, until_ts, tag_filters, limit)
+                except sqlite3.OperationalError:
+                    logger.warning("FTS5 advanced search failed, falling back", exc_info=True)
+            else:
+                logger.warning(
+                    "commands_fts missing or incomplete relative to authoritative "
+                    "commands; skipping FTS5 advanced search (#467)"
+                )
 
         # Check if FTS5 is enabled
         fts_enabled = False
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

bitflicker64 (commented):
Summary: Adds consistency checks for commands_fts and search_index before using FTS paths, falling back to standard SQL when coverage is incomplete. Matches the linked issue: detects stale indexes without rebuilding, preserves filters and OperationalError handling.

termstory/search.py: the _commands_fts_is_consistent and _search_index_is_consistent helpers look correct (docsize shadow for external content, session_id IS NOT NULL predicate for search_index). Zero-result queries stay on the healthy path.

tests/test_search.py: good coverage of healthy, missing entry, stale trigger, filters during recovery, and OperationalError cases.

CI is still running. Will recheck once green.

bitflicker64 (approved):
Summary: Gates both FTS backends on consistency checks and falls back to standard SQL when indexes are incomplete or stale. Matches #467.

termstory/search.py: _commands_fts_is_consistent uses trigger presence plus bidirectional docsize membership; _search_index_is_consistent uses per-session multiplicity via EXCEPT. OperationalError fallback preserved.

tests/test_search.py: healthy path, zero results, missing entries, dropped triggers, stale text, offsetting IDs, filters during recovery.

CI green. No unrelated changes.

bitflicker64:
Approved. Merge is blocked because main moved after #472 landed and this branch is from a fork (cannot update head from here). Please rebase onto current main and push so it can be squash-merged.

bitflicker64 (approved):
Looks good. Diff does what #467 asks: consistency gates on commands_fts and search_index skip stale indexes and fall back to _search_standard, preserves OperationalError path, no schema changes, and tests in tests/test_search.py cover healthy, missing, stale, filters, and offsetting IDs. CI green. Merging.

PR #472, opened 2026-08-27
fix(reminder): recover failed consolidation

### LINKED ISSUE #468: fix(reminder): recover failed background consolidation without losing work
## Context

Termstory's reminder/sleep-consolidation flow collects command context, groups it into clusters, generates summaries, and persists the resulting context for later use.

This is a background workflow, so individual LLM or processing failures should not invalidate unrelated work that was successfully collected.

## Problem

The consolidation pipeline currently needs stronger failure isolation and recovery semantics.

A failure while processing one cluster can potentially affect the overall consolidation operation:

1. multiple clusters are collected for consolidation;
2. one cluster reaches the LLM/request or processing stage;
3. that cluster fails due to a provider error, timeout, malformed response, or another exception;
4. the surrounding consolidation operation may fail or leave ambiguous state;
5. successfully processable clusters can therefore be lost, duplicated, or retried unnecessarily.

This is particularly problematic for a background process because the failure may not be visible to the user immediately.

## Expected

Background consolidation should treat each cluster as an independently recoverable unit.

The implementation should:

- isolate failures to the cluster that caused them;
- preserve successfully processed clusters;
- avoid marking failed work as successfully consolidated;
- make retry behavior deterministic;
- avoid duplicating persisted summaries when the same work is retried;
- preserve the existing disabled-provider behavior;
- continue processing remaining clusters after an individual failure where safe.

The fix should be based on the existing consolidation/persistence lifecycle rather than introducing a new job system.

## Acceptance criteria

- A failure processing one cluster does not prevent unrelated valid clusters from being consolidated.
- Successfully processed clusters are persisted exactly once.
- A failed cluster remains recoverable rather than being silently discarded.
- Retrying failed work does not create duplicate persisted context.
- LLM timeout/request failures are handled consistently with other cluster-processing failures.
- Existing successful consolidation behavior remains unchanged.
- Regression tests simulate multiple clusters with one deliberately failing and verify that the successful cluster is still persisted.
- Tests cover retry/idempotency behavior for the failed cluster.
- No database schema migration is introduced unless the existing persistence model genuinely requires one.

## Scope

Keep the change within the reminder/consolidation workflow and its tests.

- `termstory/reminder.py`
- persistence/database helpers used by consolidation
- existing sleep/consolidation tests
- `tests/test_reminder.py`
- `tests/test_sleep.py`

Before changing behavior, trace the complete flow from cluster collection through summary generation and persistence. Do not simply wrap the entire function in a broad exception handler, because that could hide partial-success and persistence failures.

deepmhatre13:
## Summary

Fix background sleep consolidation so a failure while generating one cluster summary does not cause successfully processed work to be lost or advance the consolidation checkpoint past failed work.

### What changed

- Isolate failures to the individual `generate_cluster_summary()` call.
- Preserve successfully processed clusters when safe to persist them.
- Prevent `MAX(end_time)` from advancing past the earliest failed cluster.
- Keep failed and post-failure commands eligible for the next consolidation run.
- Preserve idempotency across retries without introducing a database schema change.
- Keep database persistence failures outside the cluster-generation exception boundary.
- Preserve the existing disabled-provider behavior.
- Add regression coverage for partial failure, failure-first ordering, retryability, idempotency, provider timeouts, all-success behavior, and disabled-provider behavior.

### Validation

- `pytest tests/test_reminder.py --timeout=600 -q --durations=8` — 28 passed
- `pytest tests/test_sleep.py --timeout=600 -q` — 4 passed
- `git diff --check` — clean

No frontend, CLI, schema, migration, search, or unrelated changes were made.

Closes #468

### DIFF
--- a/termstory/reminder.py
+++ b/termstory/reminder.py
@@ -267,6 +267,19 @@ def generate_cluster_summary(commands: List[str]) -> str:
     containing blacklisted commands (vault, aws configure, gh auth, raw token
     strings, etc.) never reach the LLM; the standard redaction marker is returned
     instead.
+
+    Uses the configured AI provider when one is active, otherwise falls back
+    to a local, first-token-per-command summary (the ``active_provider ==
+    "disabled"`` path, which performs no network I/O and is unchanged).
+
+    Raises:
+        Exception: Whatever the configured LLM layer raises (e.g. provider,
+            request, timeout, or response-processing errors not normalized to
+            ``None`` by :func:`termstory.ai._send_llm_request`). Background
+            consolidation treats such exceptions as isolated, per-cluster
+            failures that remain recoverable on the next run (see
+            :func:`consolidate_sleep_contexts`); callers should not assume an
+            exception here means any sibling cluster failed.
     """
     from termstory.config import load_config, get_config_value
     config = load_config()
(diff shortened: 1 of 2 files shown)


bitflicker64 (approved):
Summary: Isolates generate_cluster_summary failures per cluster, persists only the chronological prefix before the earliest failure, and leaves failed work eligible via the MAX(end_time) checkpoint. Matches #468.

termstory/reminder.py ~445-528: try/except around generate_cluster_summary, cmd_min_ts bookkeeping, and the failed_clusters prefix path look correct. Persistence failures still propagate.

tests/test_reminder.py: covers partial failure, failure-first, retry, idempotency, provider timeout, all-success, and disabled provider.

CI green. No unrelated changes.

deepmhatre13:
:miku /ecsoc

PR #473, opened 2026-08-28
fix(session): recover incomplete sessions after interrupted recording

### LINKED ISSUE #469: fix(session): recover incomplete sessions after interrupted recording
## Context

Termstory records sessions incrementally while commands are executed. Session state and command records therefore represent a lifecycle rather than a single atomic write.

An interrupted process can occur at any point during recording, including during command persistence or session finalization.

## Problem

An abrupt process termination can leave session state inconsistent with the command records that were already persisted.

Potential states include:

- a session remains marked as active even though the recording process no longer exists;
- commands have been persisted but session completion metadata has not;
- the final command/session timestamps do not agree;
- duration-related fields do not reflect the commands actually stored;
- restarting Termstory can encounter an incomplete session without a deterministic recovery path.

The important issue is not merely preventing an exception. The database must remain internally coherent after an interrupted recording.

## Expected

Termstory should detect and safely recover incomplete sessions using the data already persisted.

Recovery should:

- identify sessions that were left incomplete by an interrupted recording;
- derive missing final state from authoritative persisted data where possible;
- avoid modifying genuinely active sessions incorrectly;
- remain idempotent if recovery runs more than once;
- preserve all commands already recorded;
- avoid creating duplicate sessions or commands;
- leave recovered sessions compatible with existing reporting, search, statistics, and export functionality.

## Acceptance criteria

- A simulated interruption during session recording leaves a recoverable database state.
- On the next startup/recovery operation, the incomplete session is detected and finalized consistently.
- Persisted commands are preserved exactly once.
- Session timestamps/duration remain internally consistent after recovery.
- A second recovery pass produces no additional changes.
- A genuinely completed session is not modified by the recovery logic.
- Regression tests cover interruption before and after at least one command has been persisted.
- Existing session/search/export tests continue to pass.
- No schema migration is introduced unless required by the existing data model.

## Scope

Keep the implementation within session lifecycle/recovery and its tests.

- session recording/persistence code;
- `termstory/models.py`;
- database transaction helpers;
- startup/recovery paths;
- relevant session and integration tests.

First trace the existing session lifecycle and transaction boundaries. Do not solve this by blindly marking every open session as completed; the recovery logic must distinguish an interrupted recording from a legitimately active session.

deepmhatre13:
Closes #469
## Summary

Fix incomplete persisted sessions left behind when recording is interrupted before session finalization completes.

### What changed

- Detect sessions that contain persisted commands but have a missing `end_time` or `duration_seconds`.
- Recover missing finalization fields from the authoritative persisted command timestamps.
- Preserve already-complete sessions exactly as authored.
- Leave commandless sessions untouched.
- Fill only missing fields rather than rewriting valid session metadata.
- Keep recovery idempotent and atomic within the existing database initialization transaction.
- Add regression coverage for recovery, idempotency, commandless sessions, partially missing metadata, and preservation of authored session windows.

### Validation

- `tests/test_database.py` — 25 passed
- Recovery tests — 7 passed
- Related database/session/search/archive/FTS/tag/insight/stat/predict/ask/replay/timeline/MCP tests passed
- `git diff --check` — clean

Known failures in unrelated project/symlink tests are pre-existing Windows/environment issues.

### DIFF
--- a/termstory/database.py
+++ b/termstory/database.py
@@ -253,10 +253,15 @@ class Database:
                         created_at INTEGER DEFAULT (strftime('%s', 'now'))
                     );
                 """)
-                # One-time migrations
+                # One-time migrations and consistency repairs
                 self._migrate_projects_unique_path(cursor)
                 self._migrate_deduplicate_sessions(cursor)
                 self._migrate_fts5(cursor)
+                # Recover sessions left inconsistent by an interrupted
+                # recording (#469). Runs after the schema/migration helpers so
+                # every column it touches is guaranteed to exist, and commits
+                # together with them in the single transaction below.
+                self._recover_incomplete_sessions(cursor)
                 conn.commit()
                 break
             except sqlite3.OperationalError as e:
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

bitflicker64 (approved):
Looks good. Diff does what #469 asks: _recover_incomplete_sessions fills only missing end_time/duration_seconds from MAX(commands.timestamp), leaves complete and commandless sessions alone, runs in init_db, and tests in tests/test_database.py cover recovery, idempotency, and authored windows. CI green. Merging.

PR #481, opened 2026-08-28
fix(mcp): redact local cwd from snapshots

deepmhatre13:
Closes #477
## Summary

Redact machine-specific working-directory information before persisting MCP snapshots.

## What changed

- Added `_scrub_cwd()` for persisted MCP snapshot working-directory values.
- Redacts known home-directory and username path components.
- Handles Unix and Windows-style paths, including Windows case variations.
- Preserves `cwd=None` behavior.
- Keeps the original unredacted cwd for Git/repository detection.
- Leaves existing Git-status sanitization unchanged.
- Added regression coverage for Unix, Windows, persistence, deleted cwd, and the real-cwd/Git boundary.
- No database schema or migration changes.

## Validation

- `pytest tests/test_mcp_snapshot.py -q` — 21 passed
- `pytest tests/test_replay.py -q` — 9 passed
- Relevant database/sanitizer/formatter/query tests — 71 passed
- `git diff --check` — clean

Known unrelated Windows/platform failures were reproduced against pristine `upstream/main`.

PR #485, opened 2026-08-28
fix(backup): validate restored databases before replacement

### LINKED ISSUE #479: fix(backup): validate restored SQLite databases before replacing active data
## Context

Termstory provides SQLite database backup and restore functionality. Backups are created using SQLite's backup API, and restoration currently copies the selected backup directly into the configured database path.

## Problem

`restore_db()` opens the backup and destination databases and performs the SQLite backup operation directly against the active database.

There is no validation step establishing that the selected backup is a usable Termstory database before the destination is overwritten.

A damaged, truncated, incompatible, or otherwise invalid `.db` file can therefore become the active database if SQLite accepts the connection but the database is not actually usable by Termstory.

The failure is particularly dangerous for restore operations because the operation is intended to recover data, not replace a working database with an invalid recovery source.

## Expected

Restore should validate the source database before it becomes the active database.

The implementation should:

- validate that the source is a readable SQLite database;
- verify the expected Termstory schema/tables are present;
- run an appropriate SQLite integrity check;
- restore into a temporary destination first;
- only replace the active database after validation succeeds;
- preserve the original database when validation or restore fails.

## Acceptance criteria

- Invalid/non-SQLite backup files are rejected before replacing the active database.
- A structurally invalid SQLite backup is rejected.
- A valid Termstory backup restores successfully.
- A failed restore leaves the existing active database unchanged.
- Restore remains idempotent when applied repeatedly to the same valid backup.
- Regression tests cover valid, malformed, and structurally incomplete backup files.
- No schema migration is introduced.

## Scope

Keep the change limited to backup/restore validation and atomic replacement behavior plus tests. Do not redesign the backup format.

deepmhatre13:
Closes #479

## Summary

- Validate backup readability, required Termstory schema, and SQLite integrity before restore.
- Restore into a temporary database rather than directly overwriting the active database.
- Re-validate the restored temporary database before replacement.
- Atomically replace the active database with `os.replace()`.
- Clean up temporary database and SQLite sidecar artifacts on failure.
- Handle restore validation failures cleanly in the CLI.
- Preserve the active database when validation or restoration fails.

## Validation

- `pytest tests/test_backup.py -q` — 13 passed
- `pytest tests/test_cli_restore_confirmation.py -q` — 11 passed
- `pytest tests/test_database.py -q` — 18 passed
- `pytest tests/test_save_data_none_safety.py -q` — 5 passed
- Combined — 47 passed
- `git diff --check` — clean

No schema or migration changes.

### DIFF
--- a/termstory/backup.py
+++ b/termstory/backup.py
@@ -158,27 +158,39 @@ def _validate_backup(path: str) -> None:
 
 
 def _sidecar_glob(base: str) -> list:
-    """Return existing SQLite sidecar files (``-wal``, ``-shm``, ``-journal``)
-    associated with *base* (without following into other databases)."""
+    """Return the SQLite sidecar file paths (``-wal``, ``-shm``, ``-journal``)
+    that would be associated with the exact database at *base*.  No filesystem
+    glob is performed — the three paths are derived deterministically so that
+    only sidecars belonging to this exact database path are ever considered."""
     return [f"{base}{suffix}" for suffix in ("-wal", "-shm", "-journal")]
 
 
-def _remove_sidecars(base: str) -> None:
-    """Remove SQLite sidecar files (``-wal`` / ``-shm`` / ``-journal``) for
-    the database at *base* path.  Missing files are silently skipped; removal
-    failures are logged but never raised."""
+def _remove_orphaned_sidecars(base: str) -> None:
+    """Remove SQLite sidecar files for the database at *base* path.
+
+    This is only safe to call AFTER the replacement at *base* has succeeded,
+    at which point any ``-wal``/``-shm``/``-journal`` file still present belongs
+    to the *old* (now replaced) database and would otherwise be replayed against
+    the new main file, corrupting it.  The new database is a clean single file,
+    so it has no sidecars of its own to preserve.
+
+    Raises:
+        OSError: if a stale sidecar exists but cannot be removed.  A leftover
+            stale sidecar is a correctness hazard (it can corrupt the new DB on
+            the next open), so removal failure is surfaced rather than logged
+            and silently ignored.
+    """
     for sidecar in _sidecar_glob(base):
         if os.path.exists(sidecar):
-            try:
-                os.remove(sidecar)
-            except OSError:
-                logger.warning("Could not remove stale SQLite sidecar %s", sidecar)
+            os.remove(sidecar)
 
 
 def _cleanup_temp_db(temp_db_path: str) -> None:
     """Remove a temporary database file and any of its sidecar files.
 
-    Safe to call even if the files do not exist.  Never raises.
+    Safe to call even if the files do not exist.  Best-effort by design: a
+    temporary artifact that cannot be removed is logged, never raised, so that
+    the original restore error is preserved.  Never raises.
     """
     if not temp_db_path:
         return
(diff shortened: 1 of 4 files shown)


deepmhatre13:
:miku /ecsoc

termstory/backup.py
```diff
@@ -84,23 +102,287 @@ def backup_db() -> str:
...
+        if self._fd is not None:
+            return  # already held
+        try:
+            self._fd = os.open(
+                self._lock_path, os.O_CREAT | os.O_EXCL | os.O_WRONLY, 0o666
+            )
+        except FileExistsError as exc:
+            raise BackupError(
+                "Another database restore is already in progress."
+            ) from exc
```
bitflicker64, line 229: `O_CREAT | O_EXCL` with no staleness check. A restore killed between `acquire` and `release`, by SIGKILL or a power loss, leaves `<db>.restore.lock` behind permanently, and every later restore then fails with "Another database restore is already in progress."

The sleep daemon this docstring cites does the recovery half as well: `termstory/reminder.py:560-578` writes its PID, and on `FileExistsError` reads it back, probes with `os.kill(pid, 0)`, and reclaims ownership when the owner is gone. This lock writes nothing and never probes.

Either mirror that PID probe, or drop the lock from this PR and file the concurrency hardening separately.

bitflicker64 (changes requested):
The validation and atomic-replace core of this is good work and it does what #479 asked. I have one correctness objection on the restore lock that I want fixed before this lands, plus a scope question.

## What the diff gets right

The ordering in `restore_db` is the important part and it is correct. `_validate_backup(backup_path)` at `termstory/backup.py:334` runs before anything touches the active database, the copy goes into a `tempfile.mkstemp` file in the same directory as `db_path` so `os.replace` stays on one filesystem, and the temporary copy is re-validated at `termstory/backup.py:378` before it is allowed to become the active database. The active DB is genuinely untouched on every failure path.

`_validate_backup` is right to force a `sqlite_master` read at `termstory/backup.py:127`. `sqlite3.connect` is lazy and will open any file at all, so without a query that touches the database, a text file would pass. Wrapping `PRAGMA integrity_check` in its own `except sqlite3.DatabaseError` is also correct, since on a badly corrupted file the pragma itself raises rather than returning a row.

The WAL checkpoint at `termstory/backup.py:369` before the swap, and deferring sidecar removal until after `os.replace` has succeeded, are both the right way round. A stale `-wal` replayed against the new main file would corrupt it, and removing the old sidecars any earlier would damage the live database if the replace then failed. `_sidecar_glob` deriving the three exact paths rather than globbing is a good call too.

## Blocking: the restore lock cannot recover from a stale file

`_RestoreLock.acquire` at `termstory/backup.py:219` claims the lock with `O_CREAT | O_EXCL` and has no staleness handling. If a restore is killed between `acquire` and `release`, by SIGKILL, an OOM kill, or the machine losing power, `<db>.restore.lock` is left on disk forever. Every subsequent restore then fails with "Another database restore is already in progress." and there is no way out except deleting a dotted lock file whose existence is not documented or surfaced anywhere.

That is a bad failure mode for this command in particular. Restore is what a user reaches for when something has already gone wrong, so it is exactly the operation that must not be permanently wedged by a previous crash.

The docstring says this uses the same claim as the sleep daemon PID file, but the daemon does the other half of the job. `start_sleep_daemon` in `termstory/reminder.py:560-578` writes its PID into the file, and on `FileExistsError` it reads that PID back, probes it with `os.kill(pid, 0)`, and reclaims ownership on the next attempt when the owner is gone. `_RestoreLock` writes nothing into the file and never probes, so it inherits the exclusion but not the recovery.

Two ways to close this, either is fine by me:

1. Follow the daemon properly: write `os.getpid()` into the lock, and on `FileExistsError` read it back, `os.kill(pid, 0)`, and reclaim on a stale or unparseable owner.
2. Drop `_RestoreLock` from this PR entirely. See the scope note below.

## Scope

#479 says to keep the change to backup and restore validation and atomic replacement plus tests. The cross-process lock and the `BEGIN IMMEDIATE` quiesce at `termstory/backup.py:390` are concurrency work beyond that. They are not wrong in principle, but they are the only part of this PR I have an objection to, and they are separable from the validation work that the issue actually asked for.

My preference is to cut both from this PR and file the concurrency hardening separately, where a stale-lock design can be argued on its own. `os.replace` is already atomic at the pathname level, so the validation and replacement story stands up without them. If you would rather keep the lock here, then option 1 above is what it needs.

## Smaller note

`_remove_orphaned_sidecars` is documented as raising `OSError` on failure, and it is called at `termstory/backup.py:412` after `replaced = True`. So a stale sidecar that cannot be deleted raises out of `restore_db` after the restore has actually succeeded, and `restore_cmd` reports a failure for a database that was in fact replaced. Worth either catching that and warning, or having the CLI distinguish it, so the user is not told the restore failed when it did not.

## CI

Green on 3.9 through 3.12 plus lint. My objection is a design one, not a test failure. I have not run the suite locally, so GitHub's result is the only test signal I am going on.

Happy to take this quickly once the lock question is settled.

bitflicker64 (commented):
CI green. The validate-then-atomic-replace approach for restore is sound and the tests cover the failure paths.

This is a size/XL change touching backup restore plumbing. Needs a manual look before merge.

bitflicker64 (commented):
The #479 core in termstory/backup.py is sound: validate source, copy to a temp file, re-validate, os.replace, then drop old sidecars. CLI maps BackupError to a clean exit.

Not merging. The same diff also lands session recovery in termstory/database.py, FTS consistency gates in termstory/search.py, cwd scrubbing in termstory/mcp_snapshot.py, uv language tokens in termstory/stats.py, and merged_branches ordering in termstory/git_integration.py. That is well beyond #479, and restore_db is a full rewrite of the live-database replace path. Needs a manual look. Split the unrelated work first.

bitflicker64:
Holding this for a manual look. Restore replacement is rewritten and the diff also bundles work for other issues. Split to backup/restore only before this can land.

deepmhatre13:
Cleaned up PR #485 as requested.

I separated the unrelated work from the branch and rebuilt the PR so it now contains only the #479 backup/restore changes.

The unrelated #471, #473, #481, #488, and #492 changes have been removed from the PR.

Validation:

* `tests/test_backup.py` + `tests/test_cli_restore_confirmation.py`: **31 passed**
* `git diff --check`: clean
* PR diff now contains only the intended backup/restore files.

The branch was updated using `--force-with-lease`.

termstory/backup.py
```diff
@@ -84,23 +102,287 @@ def backup_db() -> str:
     return backup_path
 
 
+def _validate_backup(path: str) -> None:
```
bitflicker64, line 105: Validating readability, required tables, and PRAGMA integrity_check before touching the live DB matches #479. The later os.replace plus sidecar cleanup is the part that actually keeps a failed restore from replacing good data.

bitflicker64 (approved):
Matches #479. restore_db now validates the backup, copies into a temp file, re-validates, then os.replace, and leaves the live DB alone on failure. CLI maps BackupError to a clean exit. CI is green.

Issue #427, opened 2026-08-15
fix: sleep daemon PID file not cleaned up on SIGTERM


bitflicker64:
fix: sleep daemon PID file not cleaned up on SIGTERM

## Bug Report

**Describe the bug**
When the sleep daemon receives SIGTERM (e.g. system shutdown, `kill <pid>`), it exits without removing `sleep_daemon.pid`. On next startup, `start_sleep_daemon()` finds a stale PID file and must go through the reclaim path unnecessarily.

**Expected behavior**
The daemon should register a SIGTERM handler that removes the PID file before exiting.

**Suggested fix**
Add `signal.signal(signal.SIGTERM, cleanup_handler)` in `run_sleep_daemon()`.

bitflicker64:
:miku /ecsoc

sanjana2505006:
:miku /claim

sanjana2505006:
This is already fixed on main. @bitflicker64 

`run_sleep_daemon()` registers `cleanup_pid` for `SIGTERM`/`SIGINT`, which removes `sleep_daemon.pid` before `sys.exit(0)`. There’s also a `finally` block that removes the PID file on other exit paths.

Landed in e327f07 (fix batch #33)

PR #486, opened 2026-08-28
test(export): cover date filter boundary consistency

### LINKED ISSUE #483: fix(export): make date-filter boundaries consistent across export paths
## Context

Termstory supports date-bounded exports, with filtering applied before records are serialized into the selected export format.

## Problem

The export path should use one consistent timestamp-boundary contract, but different filtering paths can interpret `since`/`until` boundaries differently.

This can cause records exactly on a requested boundary to be included by one export path while excluded by another, or cause equivalent date ranges to produce different record sets depending on the export format.

The filtering semantics should be traced from CLI argument parsing through the exporter and database queries and then normalized.

## Expected behavior

* `since` and `until` use one documented boundary convention.
* The same date range selects the same underlying records regardless of export format.
* Records exactly on the lower/upper boundary are handled consistently.
* Date-only inputs and timestamp inputs do not silently produce different filtering semantics.
* Existing exports without date filters remain unchanged.
* No records outside the requested interval are exported.

## Scope

Limit the change to export filtering semantics and regression tests.

Do not redesign export formats, database schema, or unrelated query APIs.

## Regression coverage

Add tests for:

* lower-boundary timestamps;
* upper-boundary timestamps;
* records exactly equal to both boundaries;
* date-only versus timestamp filters;
* equivalent filters across supported export formats;
* no-filter behavior remaining unchanged.

deepmhatre13:
Closes #483

## Summary

Add regression coverage for export date-filter boundary consistency.

## What was verified

The existing implementation already provides a single date-filtering path through `fetch_export_data()` before dispatching to JSON, CSV, and Markdown serializers.

The database query uses inclusive lower and upper bounds:

- `since` uses `>=`
- `until` uses `<=`

Date-only and explicit timestamp inputs are also handled according to the existing contract.

## Changes

- Added cross-format regression coverage proving JSON, CSV, and Markdown serialize the same filtered session set.
- Added exact `since == until` boundary coverage.
- Added explicit-time `since` boundary coverage.
- Added explicit-time `until` boundary coverage.
- Added date-only versus equivalent explicit-timestamp range coverage.

No production behavior was changed because the audit showed the existing implementation already satisfied the intended contract.

## Validation

- `tests/test_exporter.py` — 51 passed
- `tests/test_database.py` — 25 passed
- Combined exporter/database tests — 76 passed
- `git diff --check` — clean

No schema, database API, CLI, notebook, or export-format changes were made.

### DIFF
--- a/tests/test_exporter.py
+++ b/tests/test_exporter.py
@@ -876,7 +876,6 @@ def test_parse_until_preserves_explicit_time():
     assert parse_until("2026-06-10T00:00:00") == int(datetime(2026, 6, 10, 0, 0, 0).timestamp())
 
 
-
 def _make_attribution_db(path, projects, sessions, commands):
     """Create an initialised DB seeded with the given projects/sessions/commands.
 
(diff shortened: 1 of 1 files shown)


deepmhatre13:
:miku /ecsoc

bitflicker64 (commented):
Good tests, and I accept the audit conclusion, but two things keep this from merging today.

## The tests

`test_cross_format_same_session_set` at `tests/test_exporter.py:900` is the one that earns its place. Calling `fetch_export_data` once and feeding that same list to `export_json`, `export_csv`, and `export_markdown` is the correct shape for this: it pins the single choke point and would catch any serializer that started re-interpreting the date filters on its own. The five-session fixture straddling both edges, with D and E just outside, is well chosen.

`test_fetch_export_data_explicit_time_until` at `tests/test_exporter.py:962` is the one I would least like to lose. `--until 2026-06-10` expanding to end-of-day while `--until 2026-06-10T12:00:00` stays at noon is the kind of asymmetry that gets "simplified" away by someone who has not read the parser, and now it fails loudly.

`_session_ids_from_markdown` parsing the `## Session #` heading is a little tied to the markdown layout, but there is no structured markdown output to assert against, so I do not see a better option.

## The framing

The PR is titled `test(export)` and adds no production change, but #483 is filed as `fix(export): make date-filter boundaries consistent across export paths`. I am fine with the outcome, and your reasoning holds up against the code: `fetch_export_data` is the only place the bounds are applied and the query is inclusive on both ends, so the formats cannot diverge. But I want that stated plainly rather than implied, because "this issue described a bug that does not exist, here is the coverage proving it" is a different result from "fixed". I will close #483 with that note when this lands.

## Why this is not merged

`Test (Python 3.9)` and `Test (Python 3.10)` failed on the latest run:

```
FAILED tests/test_reminder.py::test_start_sleep_daemon_concurrent_invocations_spawn_once - assert 2 == 1
```

Nothing to do with this diff, which touches one test file and no production code. It is the sleep daemon PID file race in `termstory/reminder.py:560-576`, flaky across versions, related to #427.

Required checks have to be green on the commit I merge, and I am not going to hand-wave a red because I have decided it is unrelated. I also cannot run the suite locally, so CI is the only test signal I have here.

Get a green run and I will merge this as-is.

bitflicker64 (changes requested):
The export date-boundary tests look solid and match the contract described for #483.

The same PR also modifies termstory/reminder.py (PID claim grace handling) and adds related reminder tests. That is unrelated to export date filters. Please move the reminder changes to a separate PR so this one stays scoped to the export coverage.

bitflicker64 (changes requested):
The tests in tests/test_exporter.py match the #483 contract: one fetch_export_data result fed to JSON/CSV/Markdown, inclusive since==until, explicit-time since/until, date-only vs explicit range.

Cannot merge: required checks are red. Latest run: Test (Python 3.9) failed, Test (Python 3.11) failed. Lint and 3.10/3.12 passed.

Get a fully green run.

tests/test_exporter.py
```diff
@@ -875,3 +875,118 @@ def test_parse_until_preserves_explicit_time():
...
+    db = _build_session_db(tmp_path / "xfmt.db", [
+        (datetime(2026, 6, 9, 23, 59).timestamp(), None),        # D: just before
+        (datetime(2026, 6, 10, 0, 0).timestamp(), None),         # A: exact lower
+        (datetime(2026, 6, 10, 12, 0).timestamp(), None),        # B: strictly inside
+        (datetime(2026, 6, 10, 23, 59, 59).timestamp(), None),   # C: exact upper
+        (datetime(2026, 6, 11, 0, 0).timestamp(), None),         # E: just after
+    ])
+
+    # Single choke-point: the one filtered session list every format consumes.
+    sessions = fetch_export_data(db, since_str="2026-06-10", until_str="2026-06-10")
```
bitflicker64, line 1199: These tests document the current shared fetch_export_data path. They do not change production code. CI is not green (Python 3.9 and 3.11 failed on the latest run), so this cannot land yet.

bitflicker64 (changes requested):
The tests describe the current inclusive since/until contract across JSON/CSV/Markdown. No production change, which is fine if the audit is correct. Cannot merge: latest CI failed on Python 3.9 and 3.11, mergeable_state is unstable.

bitflicker64 (commented):
Summary: tests-only coverage for export date bounds in `tests/test_exporter.py`. The cases (cross-format session set, since==until, explicit since/until) match the #483 contract, and the PR does not change production code.

Blocked: mergeable_state is dirty (conflicts with main), and required checks Test (Python 3.9) and Test (Python 3.11) failed on the latest run. Rebase onto current main and wait for a full green suite.

tests/test_exporter.py
```diff
@@ -875,3 +875,118 @@ def test_parse_until_preserves_explicit_time():
...
+    expected = {2, 3, 4}  # A, B, C
+
+    export_json(sessions, db, output_file=None)
```
bitflicker64, line 1202: This proves the three serializers emit the same IDs from one already-filtered list. It does not prove each CLI export path applies the same date bounds. That is fine as coverage, but Test (Python 3.9) and Test (Python 3.11) failed on the latest check run so this cannot merge yet.

bitflicker64 (changes requested):
Summary: tests only, no production change. The new cases lock inclusive since/until bounds and same session IDs across JSON/CSV/Markdown when they share one `fetch_export_data` result. That matches the audit in the PR body for #483.

Cannot merge: Test (Python 3.9) and Test (Python 3.11) failed. mergeable_state=unknown.

CI: Lint pass, Test 3.10/3.12 pass, Test 3.9/3.11 fail.

tests/test_exporter.py
```diff
@@ -875,3 +875,118 @@ def test_parse_until_preserves_explicit_time():
...
+def _session_ids_from_markdown(text):
+    ids = []
+    for line in text.splitlines():
+        if line.startswith("## Session #"):
+            # Heading is "## Session #<id> — <project>".
+            ids.append(int(line.split("## Session #", 1)[1].split(" ", 1)[0]))
+    return sorted(ids)
+
+
+def test_cross_format_same_session_set(tmp_path, capsys):
```
bitflicker64, line 1182: Good check that JSON/CSV/Markdown all consume the same `fetch_export_data` result. This is what #483 asked for: one filter path, inclusive bounds, no per-format reinterpretation.

tests/test_exporter.py
```diff
@@ -875,3 +875,118 @@ def test_parse_until_preserves_explicit_time():
...
+    """--until with an explicit time must NOT expand to end-of-day."""
+    db = _build_session_db(tmp_path / "explicit_until.db", [
+        (datetime(2026, 6, 10, 12, 0, 0).timestamp(), None),  # exact  -> included
+        (datetime(2026, 6, 10, 12, 0, 1).timestamp(), None),  # after  -> excluded
+        (datetime(2026, 6, 10, 12, 0, 2).timestamp(), None),  # after  -> excluded
+    ])
+    sessions = fetch_export_data(db, until_str="2026-06-10T12:00:00")
+    assert [s.id for s in sessions] == [1]
+
+
```
bitflicker64, line 1256: Useful contract test: date-only `2026-06-10` equals the explicit `00:00:00` .. `23:59:59` window. Keep this if the parse helpers ever change again.

bitflicker64 (commented):
Summary: tests-only coverage for export date bounds. Confirms JSON/CSV/Markdown share `fetch_export_data()`, inclusive since/until, and date-only vs explicit timestamp equivalence. Matches #483 without production changes.

CI is not green: latest Test (Python 3.9) and Test (Python 3.11) failed.

Not merging until those jobs pass.

bitflicker64 (changes requested):
Summary: pure test coverage for export date boundaries. Matches the spirit of #483 (regression proof that filters are consistent).

CI has failures on multiple Python versions, and mergeable_state is dirty (conflicts with main). Needs a rebase and green CI.

bitflicker64 (changes requested):
Summary: tests/test_exporter.py adds cross-format and boundary cases for fetch_export_data. No production change. That documents the current contract; it does not change export paths.

tests/test_exporter.py test_cross_format_same_session_set: the markdown helper parses "## Session #<id> " by splitting on the first space after the id. If the heading format drifts, this test will pass or fail independently of the date filter.

Linked #483 asked for consistent date-filter boundaries across export paths. This PR only adds tests and states the existing path already matches.

Cannot merge: mergeable_state=dirty (conflicts with main). Latest Test (Python 3.9) and Test (Python 3.11) failed. Lint and 3.10/3.12 passed on the later run.

Rebase onto main and get every required check green.

bitflicker64 (approved):
Solid regression coverage for the export date boundaries in #483. The tests exercise the single fetch_export_data path across JSON/CSV/Markdown, exact since==until, explicit times, and date-only equivalence. No production code changes needed and none present. CI green, matches the issue.

tests/test_exporter.py
```diff
@@ -1155,3 +1154,118 @@ def test_project_filter_mixed_session_matches_named_and_other(tmp_path):
...
+    # A/B/C inside [2026-06-10 00:00:00, 2026-06-10 23:59:59]; D/E outside.
+    db = _build_session_db(tmp_path / "xfmt.db", [
+        (datetime(2026, 6, 9, 23, 59).timestamp(), None),        # D: just before
+        (datetime(2026, 6, 10, 0, 0).timestamp(), None),         # A: exact lower
+        (datetime(2026, 6, 10, 12, 0).timestamp(), None),        # B: strictly inside
+        (datetime(2026, 6, 10, 23, 59, 59).timestamp(), None),   # C: exact upper
+        (datetime(2026, 6, 11, 0, 0).timestamp(), None),         # E: just after
+    ])
+
+    # Single choke-point: the one filtered session list every format consumes.
```
bitflicker64, line 1198: Calling `fetch_export_data` once and feeding that list to JSON/CSV/Markdown is the right way to prove the formats do not re-interpret the date window.

tests/test_exporter.py
```diff
@@ -1155,3 +1154,118 @@ def test_project_filter_mixed_session_matches_named_and_other(tmp_path):
...
+    db = _build_session_db(tmp_path / "explicit_until.db", [
+        (datetime(2026, 6, 10, 12, 0, 0).timestamp(), None),  # exact  -> included
+        (datetime(2026, 6, 10, 12, 0, 1).timestamp(), None),  # after  -> excluded
+        (datetime(2026, 6, 10, 12, 0, 2).timestamp(), None),  # after  -> excluded
+    ])
+    sessions = fetch_export_data(db, until_str="2026-06-10T12:00:00")
+    assert [s.id for s in sessions] == [1]
+
+
+def test_date_only_range_equals_explicit_range(tmp_path):
```
bitflicker64, line 1257: This is the contract #483 asked for: date-only `2026-06-10` equals `00:00:00` through `23:59:59`. Good coverage of both edges.

bitflicker64 (commented):
Matches #483 as regression coverage. `fetch_export_data` is the single filter path; JSON/CSV/Markdown all emit the same session set. Inclusive since/until, since==until, explicit times, and date-only vs explicit range are covered. No production change, which is correct if the contract already holds.

Not merging until required CI is fully green and the branch is clean against main.

PR #487, opened 2026-08-28
fix(git): resolve repository identity for linked worktrees

### LINKED ISSUE #484: fix(git): prevent repository detection from attributing commands to the wrong worktree
## Context

Termstory derives Git repository/project information from the command working directory and Git metadata.

## Problem

Repository detection can become ambiguous when the working directory is inside nested repositories, linked worktrees, or paths that resolve differently through symlinks/relative components.

If repository identity is inferred from an insufficiently normalized path or from a parent repository without confirming the actual Git worktree boundary, commands can be attributed to the wrong repository.

This is particularly problematic for projects containing nested Git repositories or linked worktrees.

## Expected behavior

* Git repository detection resolves the actual repository/worktree associated with the command cwd.
* Nested repositories are not incorrectly attributed to their parent repository.
* Linked worktrees retain their correct repository/worktree identity.
* Relative path components are normalized before repository comparison.
* Detection failures remain non-fatal and preserve existing fallback behavior.
* Existing project attribution behavior remains unchanged for ordinary repositories.

## Scope

Keep the change limited to Git repository/worktree identity resolution and tests.

Do not modify command persistence, session attribution schema, search behavior, or unrelated Git commands.

## Regression coverage

Add deterministic tests for:

* normal repository;
* nested repository;
* linked worktree;
* normalized equivalent paths;
* non-repository directory;
* Git command failure/fallback behavior.

deepmhatre13:
Closes #484
## Summary

Fix Git repository/worktree identity resolution for linked worktrees and
normalized paths.

### Changes

- Recognize `.git` files used by linked Git worktrees.
- Add authoritative Git-backed repository root resolution.
- Resolve repository identity from the nearest repository/worktree.
- Normalize timestamp-detective Git-root resolution through real paths.
- Normalize project-root cache keys so equivalent paths share identity.
- Preserve existing network-mount and project detection behavior.
- Add regression coverage for ordinary repos, nested repos, linked worktrees,
  normalized paths, symlinks, failures, and cache isolation.

### Validation

All #484-specific tests pass.

The 8 failures remaining in the broader project/timestamp test suites are
pre-existing Windows/environment failures reproduced against pristine
upstream/main and are not introduced by this change.

### DIFF
--- a/termstory/git_integration.py
+++ b/termstory/git_integration.py
@@ -1,7 +1,7 @@
 import os
 import re
 import subprocess
-from typing import List, Dict
+from typing import List, Dict, Optional
 
 def is_git_repo(path: str) -> bool:
     """Check if the directory path is a valid git repository worktree"""
@@ -25,6 +25,56 @@ def is_git_repo(path: str) -> bool:
         return res.returncode == 0 and res.stdout.strip() == "true"
     except Exception:
         return False
+
+
+def find_git_root(path: str) -> Optional[str]:
+    """Authoritatively resolve the Git repository/worktree root for *path*.
+
+    Delegates to Git itself via ``git -C <path> rev-parse --show-toplevel``,
+    which correctly handles:
+
+    * ordinary repositories (``.git`` directory)
+    * linked worktrees (``.git`` file)
+    * nested repositories (inner-most repo wins)
+    * symlinked / ``../`` paths (Git normalises internally)
+
+    Returns the canonical top-level directory of the worktree, or ``None``
+    when *path* is not inside a Git repository, Git is unavailable, the
+    command times out, or the path does not exist.
+
+    Mirrors the conventions of :func:`is_git_repo` (10 s timeout,
+    ``check=False``, ``errors="replace"``).
+    """
+    abs_path = os.path.abspath(os.path.expanduser(path))
+    if not os.path.exists(abs_path) or not os.path.isdir(abs_path):
+        return None
+    try:
+        res = subprocess.run(
+            ["git", "-C", abs_path, "rev-parse", "--show-toplevel"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            errors="replace",
+            check=False,
+            timeout=10
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        # OSError covers a missing `git` executable (FileNotFoundError),
+        # permission errors, and invalid paths; TimeoutExpired covers a
+        # hung `git` in a slow filesystem. Any of these is a non-fatal,
+        # expected resolution failure, so return None.
+        return None
+
+    if res.returncode == 0:
+        root = res.stdout.strip()
+        if root:
+            # Git reports paths with forward slashes even on Windows
+            # (e.g. `D:/repo`), while os.path / Path use the platform
+            # separator.  Normalise so caller comparisons against native
+            # paths (from os.path.realpath etc.) are consistent (#484).
+            return os.path.normpath(root)
+    return None
+
 
 def clean_commit_message(message: str) -> str:
     """Clean commit message for display by removing emojis, JIRA codes, PR numbers, and prefixes"""
--- a/termstory/timestamp_detective.py
+++ b/termstory/timestamp_detective.py
@@ -199,18 +199,19 @@ class TimestampDetective:
         A two-stage strategy is used:
 
         1. Fast filesystem heuristic — walk upward from the resolved path
-           (symlinks collapsed via ``os.path.realpath``) looking for a ``.git``
-           entry.  ``os.path.exists`` is used so that **both** forms are
-           recognised:
+           (symlinks collapsed via ``os.path.realpath``).  A ``.git`` entry is
+           accepted as a repository root only when it is a genuine Git marker:
 
-           * ``.git`` **directory**  — ordinary repository
-           * ``.git`` **file**       — linked worktree (created by
-             ``git worktree add``)
+           * ``.git`` **directory**  — ordinary repository.
+           * ``.git`` **file**       — a valid linked-worktree marker (created
+             by ``git worktree add``).  The file must be a real ``gitdir:``
+             pointer whose target git directory exists; an arbitrary, malformed
+             or stale file named ``.git`` is NOT accepted, and the walk continues
+             upward so an enclosing repository can still be found.
 
         2. Authoritative Git fallback — when the heuristic fails (e.g. the
-           ``.git`` marker lives more than 10 levels up, or git is configured
-           via ``GIT_DIR``), delegate to
-           ``git -C <path> rev-parse --show-toplevel``.
+           marker lives more than 10 levels up, or git is configured via
+           ``GIT_DIR``), delegate to ``git -C <path> rev-parse --show-toplevel``.
 
         Stops after 10 levels to avoid climbing to the filesystem root on
         misconfigured machines.  Returns ``None`` if no git root is found.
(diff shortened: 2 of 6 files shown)


deepmhatre13:
:miku /ecsoc

termstory/timestamp_detective.py
```diff
@@ -194,20 +194,87 @@ def _get_npm_global_root(self) -> Optional[str]:
...
+            return False
+        try:
+            with open(git_path, "r", encoding="utf-8", errors="replace") as fh:
+                first_line = fh.readline()
+        except OSError:
+            return False  # unreadable marker -> treat as invalid
+        if not first_line.startswith("gitdir:"):
+            return False
+        target = first_line[len("gitdir:"):].strip()
+        target = first_line[len("gitdir:"):].strip()
```
bitflicker64: Duplicate assignment, line 265 does the same thing.

```suggestion
```

termstory/timestamp_detective.py
```diff
@@ -194,20 +194,87 @@ def _get_npm_global_root(self) -> Optional[str]:
...
+            return False
+        if not os.path.isabs(target):
+            # Git writes absolute targets, but resolve a relative target the
+            # way Git does: relative to the directory holding the `.git` file.
+            target = os.path.join(os.path.dirname(git_path), target)
+        return os.path.isdir(target)
+
+        if not target:
+            return False
+        return os.path.isdir(target)
```
bitflicker64: Unreachable. Line 273 returns, so nothing here can run. Looks like a leftover from an earlier version of the function.

```suggestion
```

bitflicker64 (commented):
Solid work on the identity resolution, and CI is green, but this one is not getting merged on a single pass. It introduces a new `git` subprocess call, which under my own rules means it gets a deliberate second look rather than a same-day merge. There are also two leftovers in `_is_git_marker` that need cleaning up regardless.

## Needs fixing

`termstory/timestamp_detective.py:265-266` has the same assignment twice:

```python
target = first_line[len("gitdir:"):].strip()
target = first_line[len("gitdir:"):].strip()
```

Harmless, but it is an obvious paste artifact.

`termstory/timestamp_detective.py:275-277` is unreachable. The function returns at line 273, so the trailing `if not target: return False` / `return os.path.isdir(target)` can never run. It looks like an earlier draft that did not get removed, and it reads as though there is a fallback path when there is not.

Both are inline below.

## What the change gets right

`_is_git_marker` is the right idea. Treating a `.git` file as a worktree marker only when it is a real `gitdir:` pointer whose target exists, and continuing the upward walk otherwise, means a stale marker left behind by a removed worktree cannot mask an enclosing repository. Resolving a relative `gitdir:` target against the directory holding the `.git` file matches what Git does. The walk switching from `os.path.abspath` to `os.path.realpath` at `termstory/timestamp_detective.py:214` is what makes symlinked paths land on the same root.

The cache key change in `termstory/project.py:178` is the subtlest part of the diff and I think it is correct. Keying on realpath alone would have aliased two raw spellings that `_find_project_root_impl` genuinely treats differently, because its UNC guard inspects the raw `\\` or `//` prefix before any normalisation. Carrying `raw_unc` in the tuple keeps those apart while letting `a/child` and `a/child/../child` share one entry. The comment explaining why is worth keeping.

`find_git_root` is written carefully: `check=False`, a 10 second timeout, `errors="replace"`, `OSError` catching a missing `git` binary, and `os.path.normpath` on the result because `rev-parse` reports forward slashes even on Windows. That last one matters for the comparisons against `os.path.realpath` output that this PR relies on.

## Why it is escalated rather than merged

`termstory/git_integration.py:52` adds a new `subprocess.run(["git", "-C", abs_path, "rev-parse", "--show-toplevel"])`. Anything in this repo that starts shelling out gets a manual review before it lands, no matter how green the checks are, and that rule does not bend for a call that is well written.

The thing I specifically want to look at is cost. `_find_git_root` now falls through to `find_git_root` on every miss, and `_find_git_root` sits on the timestamp detective's hot path across large legacy history parses. Before this change a miss was a bounded 10-level filesystem walk. After it, a miss is that walk plus a process spawn, and the common case for a non-repository directory is a guaranteed miss, so the fallback fires exactly where it is least useful. Over a few thousand distinct paths that is a lot of `git` invocations. I want to check whether the fallback needs a negative-result cache, or whether it should be gated to the case where a `.git` marker was seen but not resolved, before I take it.

To be clear, none of that is me saying the fallback is wrong. It is the right way to handle `GIT_DIR` and deep markers. I just want to measure it rather than assume.

## CI

Green on 3.9 through 3.12 plus lint. I have not run the suite locally, so that is the only test signal I am going on.

Next step: push the two cleanups, and I will come back to the subprocess question with the parse-path cost in hand.

bitflicker64 (commented):
CI green. The changes to recognize .git files for linked worktrees and normalize path identity look appropriate for #484. Regression coverage is thorough.

No correctness objections from the diff. Ready for merge once any outstanding review comments are addressed.

bitflicker64 (commented):
Addresses #484. _is_git_marker accepts a gitdir: worktree file, project cache keys on (raw_unc, realpath), and find_git_root uses git rev-parse --show-toplevel. Tests cover ordinary repos, nested repos, linked worktrees, stale markers, and cache isolation.

Not merging. termstory/git_integration.py find_git_root adds a new subprocess.run(["git", "-C", abs_path, "rev-parse", "--show-toplevel"]). Needs a manual look.

bitflicker64:
Holding this for a manual look. find_git_root adds a new git subprocess call on the identity-resolution path.

termstory/git_integration.py
```diff
@@ -26,6 +26,56 @@ def is_git_repo(path: str) -> bool:
...
+
+    Mirrors the conventions of :func:`is_git_repo` (10 s timeout,
+    ``check=False``, ``errors="replace"``).
+    """
+    abs_path = os.path.abspath(os.path.expanduser(path))
+    if not os.path.exists(abs_path) or not os.path.isdir(abs_path):
+        return None
+    try:
+        res = subprocess.run(
+            ["git", "-C", abs_path, "rev-parse", "--show-toplevel"],
```
bitflicker64, line 53: New subprocess.run for git rev-parse --show-toplevel. Worktree identity is a real bug, but a new git subprocess from project/timestamp paths needs a manual look before merge. The filesystem .git file heuristic in timestamp_detective.py is the safer half of this change.

bitflicker64 (commented):
Needs a manual look. find_git_root adds a new git subprocess (rev-parse --show-toplevel) and timestamp_detective falls back to it. That is the right idea for linked worktrees (#484), but I am not merging a new subprocess path from here. CI is green. Please wait for a maintainer pass on the git invocation and the cache-key change in project.py.

bitflicker64:
Needs a manual look before merge. This adds a new git subprocess in find_git_root and wires it into timestamp detection. Holding it here.

bitflicker64 (commented):
Summary: this adds `find_git_root()` in `termstory/git_integration.py` via `git rev-parse --show-toplevel`, teaches `timestamp_detective._find_git_root` to accept `.git` files, and changes project-root cache keys in `termstory/project.py`. That is aimed at #484.

Not merging from here. New subprocess call plus a large rewrite across git/project/timestamp_detective needs a manual look. Tests also spawn real `git init` / `git worktree add` processes.

CI on the current head is green. Please keep this scoped to worktree identity and wait for a maintainer pass.

termstory/git_integration.py
bitflicker64, line 53: New `subprocess.run(["git", "-C", abs_path, "rev-parse", "--show-toplevel"])` on a path resolution hot path. Holding this for a manual look: new subprocess call, and mergeable_state is unknown even though the recorded checks look green.

bitflicker64 (commented):
Summary: worktree identity via a `.git` file plus `git rev-parse --show-toplevel` is what #484 asked for. Tests cover ordinary repos, nested repos, linked worktrees, and stale markers.

Needs a manual look. This adds a new subprocess call on the project-root path, and mergeable_state is unknown. Not merging.

CI recorded: Lint pass, Test 3.9/3.10/3.11/3.12 pass. mergeable_state=unknown.

bitflicker64:
Needs a manual look. New git subprocess on the identity path, and mergeable_state is unknown. Not merging from this pass.

PR #488, opened 2026-08-28
[Git] Make merged_branches order deterministic

### LINKED ISSUE #448: fix(git_integration): merged_branches ordering is nondeterministic across runs
## Context

`get_timeframe_git_stats()` in `termstory/git_integration.py:118` collects merged branch names across a list of project paths and returns them at `termstory/git_integration.py:197`:

```python
"merged_branches": list(set(merged_branches))
```

## Problem

`set` iteration order for strings depends on per-process hash randomization, so the same repository state produces a different branch order on every run:

```
$ for i in 1 2 3 4; do python3 -c "print(list(set(['feature/login','bugfix/auth','release/v2','hotfix/db','chore/deps'])))"; done
['hotfix/db', 'chore/deps', 'bugfix/auth', 'release/v2', 'feature/login']
['bugfix/auth', 'chore/deps', 'hotfix/db', 'feature/login', 'release/v2']
['feature/login', 'chore/deps', 'hotfix/db', 'bugfix/auth', 'release/v2']
['chore/deps', 'release/v2', 'bugfix/auth', 'hotfix/db', 'feature/login']
```

Two consequences: rendered reports reshuffle their branch list between runs for unchanged data, and any test asserting on this field is flaky unless it sorts first.

## Expected

Deterministic ordering. Either preserve first-seen order while removing duplicates (`dict.fromkeys`), or sort the result. First-seen order is closer to the existing intent, since branches are collected newest-first from `git log`.

## Acceptance criteria

- [ ] `merged_branches` is deterministic for identical input
- [ ] Duplicates across multiple project paths are still collapsed
- [ ] Test added to `tests/test_git_integration.py` asserting exact list equality

## Scope

One line plus a test. Roughly 1 file, 6 lines.

sanjana2505006:
## Summary
* Replace `list(set(merged_branches))` with `list(dict.fromkeys(merged_branches))` so branch order is stable across runs
* Keep first-seen order (matches newest-first `git log` collection) while still collapsing duplicates across project paths
* Add an exact-equality regression test in `tests/test_git_integration.py`

Fixes #448

## Test plan
* [ ] `pytest tests/test_git_integration.py -q`
* [ ] Confirm `merged_branches` order is identical across repeated calls for the same input

### DIFF
--- a/termstory/git_integration.py
+++ b/termstory/git_integration.py
@@ -200,6 +200,6 @@ def get_timeframe_git_stats(project_paths: List[str], since_ts: int, until_ts: i
     return {
         "additions": total_additions,
         "deletions": total_deletions,
-        "merged_branches": list(set(merged_branches))
+        "merged_branches": list(dict.fromkeys(merged_branches))
     }
 
--- a/tests/test_git_integration.py
+++ b/tests/test_git_integration.py
@@ -1,6 +1,13 @@
 import os
 import subprocess
-from termstory.git_integration import clean_commit_message, is_git_repo, get_project_commits
+from unittest.mock import patch
+
+from termstory.git_integration import (
+    clean_commit_message,
+    get_project_commits,
+    get_timeframe_git_stats,
+    is_git_repo,
+)
 
 def test_clean_commit_message():
     # Test conventional commit prefix stripping
@@ -83,7 +90,6 @@ def test_is_git_repo_worktree_vs_git_dir(tmp_path):
     # The .git directory is NOT a worktree (rev-parse exits 0 but prints "false")
     assert is_git_repo(str(repo_path / ".git")) is False
 
-from unittest.mock import patch
 def test_git_missing_or_failing(tmp_path):
     # Test subprocess.run raising an exception (e.g. git not found)
     with patch("termstory.git_integration.subprocess.run") as mock_run:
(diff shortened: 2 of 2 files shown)


bitflicker64 (approved):
This is the right fix for #448.

`termstory/git_integration.py:203` swaps `list(set(merged_branches))` for `list(dict.fromkeys(merged_branches))`. That is the correct idiom here: it collapses duplicates the same way `set` did, but preserves first-seen insertion order, so the output stops depending on string hash ordering across runs.

First-seen order is also the meaningful order in this function. `merged_branches` is appended to inside the loop over `project_paths` at `termstory/git_integration.py:185-195`, which reads `git log --merges` output newest first, so the resulting list is newest-first per repository and stable across repositories. Deduping across the two paths still works, which the test covers.

On the test at `tests/test_git_integration.py:135`: patching `is_git_repo` alongside `subprocess.run` is necessary, since `get_timeframe_git_stats` skips non-repo paths before it ever runs `git log`. The `fake_run` gate on `"--merges" in cmd` is a clean way to let the numstat call return empty while the merge call returns the fixture. Asserting exact list equality rather than a set is the whole point of the regression, so that is right.

Moving the `from unittest.mock import patch` import from mid-file up to the header at `tests/test_git_integration.py:3` is a tidy-up rather than a behaviour change, and it is in a file this PR already touches, so I have no objection to it riding along.

CI is green across 3.9 through 3.12 plus lint.

Approving.

PR #489, opened 2026-08-28
[Parser] Honor TERMSTORY_DATE_OVERRIDE in bash timestamp bounds

### LINKED ISSUE #450: fix(parser): bash timestamp bounds use time.time() and ignore TERMSTORY_DATE_OVERRIDE
## Context

`termstory/parser.py` `_assign_missing_timestamps_fallback` (used by bash, fish, and powershell parsers) computes the upper bound for valid timestamps with `import time; now_clean = int(time.time())` and later `now = int(time.time())`. Zsh parsing goes through `get_current_time()` from `date_utils`.

Issue #407 closed residual `datetime.now()` call sites, but these `time.time()` paths remain.

## Problem

When `TERMSTORY_DATE_OVERRIDE` is set (tests and deterministic runs), bash/fish/powershell history filtering still uses wall-clock time. Commands whose recovered timestamps fall between the override and real now are dropped or kept inconsistently compared to the zsh path.

## Expected

All upper/lower timestamp bounds in `_assign_missing_timestamps_fallback` should use `get_current_time().timestamp()` so override is honoured the same way as the zsh path.

## Acceptance criteria

- [ ] Both `time.time()` sites in `_assign_missing_timestamps_fallback` are replaced with `get_current_time()`
- [ ] Test in `tests/test_parser.py` sets `TERMSTORY_DATE_OVERRIDE`, parses a bash history fixture, and asserts no command is filtered solely because of wall-clock skew

## Scope

1 file (`termstory/parser.py`) plus one test; roughly 10-20 lines.

sanjana2505006:
## Summary
* Replace both `time.time()` calls in `_assign_missing_timestamps_fallback` with `get_current_time().timestamp()`
* Align bash/fish/powershell timestamp bounds with the zsh path under `TERMSTORY_DATE_OVERRIDE`
* Add a regression test that keeps bash timestamps when wall-clock `time.time()` would treat them as future

Fixes #450

## Test plan
* [ ] Confirm `test_bash_history_honors_date_override_not_wall_clock` passes

### DIFF
--- a/termstory/parser.py
+++ b/termstory/parser.py
@@ -662,8 +662,7 @@ def _assign_missing_timestamps_fallback(
     commands_to_return = []
     
     # Pre-clean raw timestamps: discard any that are out of bounds
-    import time
-    now_clean = int(time.time())
+    now_clean = int(get_current_time().timestamp())
     from termstory.config import load_config
     max_history_age_clean = load_config().get("max_history_age", 5)
     five_years_ago_clean = now_clean - (max_history_age_clean * 365 * 24 * 60 * 60)
@@ -768,8 +767,7 @@ def _assign_missing_timestamps_fallback(
         else:
             first_known_timestamp = resolved_timestamps[first_known_idx]
             
-        import time
-        now = int(time.time())
+        now = int(get_current_time().timestamp())
         
         # Bounded Interpolation
         idx = 0
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -467,6 +467,27 @@ def test_assign_missing_timestamps_fallback_clamping():
     assert commands[0].timestamp == 1748851190
 
 
+def test_bash_history_honors_date_override_not_wall_clock(tmp_path, monkeypatch):
+    """Issue #450: bash timestamp bounds must follow TERMSTORY_DATE_OVERRIDE."""
+    from datetime import datetime
+
+    monkeypatch.setenv("TERMSTORY_DATE_OVERRIDE", "2027-06-01 12:00:00")
+    monkeypatch.setattr("time.time", lambda: 1_700_000_000)
+
+    override_ts = int(datetime(2027, 6, 1, 12, 0, 0).timestamp())
+    ts1 = override_ts - 120
+    ts2 = override_ts - 60
+    temp_file = tmp_path / "bash_date_override_test"
+    temp_file.write_text(f"#{ts1}\ngit status\n#{ts2}\ndocker ps\n")
+
+    commands = parse_bash_history(str(temp_file))
+    assert len(commands) == 2
+    assert commands[0].timestamp == ts1
+    assert commands[0].command == "git status"
+    assert commands[1].timestamp == ts2
+    assert commands[1].command == "docker ps"
+
+
 def test_parse_zsh_history_getmtime_oserror_falls_back(tmp_path):
     from unittest.mock import patch
 


termstory/parser.py
```diff
@@ -561,16 +561,19 @@ def _assign_missing_timestamps_fallback(
...
     five_years_ago_clean = now_clean - (max_history_age_clean * 365 * 24 * 60 * 60)
     
     cleaned_temp_commands = []
     for t, cmd in temp_commands:
         if t is not None:
-            if not (five_years_ago_clean <= t <= now_clean):
+            if t > now_clean:
+                # After effective "now" (incl. TERMSTORY_DATE_OVERRIDE): drop.
+                # Nulling would let interpolation invent an in-range timestamp.
+                continue
```
bitflicker64: This `continue` drops the command from `temp_commands` altogether. On main the same case set `t = None` and the command survived with an interpolated timestamp.

It is also asymmetric with the branch immediately below: `t < five_years_ago_clean` still nulls and keeps the command, while `t > now_clean` deletes it.

Suggest restoring the single combined bounds check so this stays a pure timekeeping fix:

```suggestion
            if not (five_years_ago_clean <= t <= now_clean):
                t = None
```

bitflicker64 (changes requested):
The two `time.time()` replacements are exactly right, but the PR also changes what happens to an out-of-range command, and that part I cannot take as written.

## The part that is correct

`termstory/parser.py:564` and `termstory/parser.py:672` swap `int(time.time())` for `int(get_current_time().timestamp())`. That is precisely what #450 asked for, it brings the bash/fish/powershell bounds in line with the zsh path, and `get_current_time` is already imported at module scope (`termstory/parser.py:8`), so the two local `import time` lines going away is a genuine cleanup rather than a lost dependency.

`test_bash_history_honors_date_override_not_wall_clock` is a good regression. Setting `TERMSTORY_DATE_OVERRIDE` to 2027 while pinning `time.time` to a 2023 value means the test fails on the old code and passes on the new, which is what a regression test for this issue should do.

## The part that needs to change

On main, an out-of-bounds timestamp sets `t = None` and the command stays in `temp_commands`. It then picks up an interpolated timestamp further down and still reaches the user's history. This PR changes the upper-bound branch to `continue`, which removes the `(t, cmd)` tuple entirely, so the command is dropped from the parse result.

That is silent data loss on a real path. A machine with clock skew, a history file synced from a host whose clock ran ahead, or a `TERMSTORY_DATE_OVERRIDE` set to a past date for a deterministic run will now have those commands vanish from `termstory` output rather than appear with an approximate timestamp. `test_bash_history_past_override_rejects_future_timestamps` locks that behaviour in as intended, which is what makes it a change in contract rather than an implementation detail.

It also leaves the two bounds inconsistent with each other: too-old still nulls and interpolates, too-new deletes.

Worth noting that the zsh path does not do this either. It nulls out-of-range values the same way, and only filters at the very end of `parse_zsh_history` after timestamps have been resolved.

#450 asks for the bounds to honour the override and for no command to be filtered solely because of wall-clock skew. Dropping commands outright is the opposite of that second point, and the issue scopes the change at roughly 10 to 20 lines.

## What I would like

Keep both branches as `t = None`, so the diff in `_assign_missing_timestamps_fallback` is just the two `get_current_time()` swaps:

```python
if not (five_years_ago_clean <= t <= now_clean):
    t = None
```

Then drop `test_bash_history_past_override_rejects_future_timestamps`, or rewrite it to assert the command survives with a re-derived in-range timestamp instead of asserting it disappears.

If you think future-dated commands genuinely should be discarded rather than re-placed, that is a reasonable argument, but it is a separate behaviour change worth its own issue so it can be judged on its own merits rather than landing inside a timekeeping fix.

CI is green on all four Python versions, so this is a design objection, not a test failure.

bitflicker64 (commented):
The change to use get_current_time().timestamp() in the fallback path is the right fix for #450 and the new test correctly guards against wall-clock drift under TERMSTORY_DATE_OVERRIDE.

CI is not fully green: Test (Python 3.11) failed. Needs a green run before merge.

bitflicker64 (changes requested):
The production change is what #450 asked for. Both time.time() calls in _assign_missing_timestamps_fallback now use get_current_time().timestamp(). test_bash_history_honors_date_override_not_wall_clock pins the override vs wall-clock case.

Cannot merge: Test (Python 3.11) failed. Re-run or fix until every required check is green.

CI: Lint pass, Test 3.9/3.10/3.12 pass, Test 3.11 fail.

termstory/reminder.py
```diff
@@ -563,7 +563,11 @@ def start_sleep_daemon(db_path: str):
             # is (or is about to become) the running daemon — defer to it.
             try:
                 with open(pid_file, "r") as f:
-                    pid = int(f.read().strip())
+                    content = f.read().strip()
```
bitflicker64: This reminder PID-file change is unrelated to #450 and to the parser date-override fix. Please drop it from this PR. Also Test (Python 3.10) is failing.

bitflicker64 (changes requested):
The parser change (time.time() -> get_current_time()) is the #450 fix. Two problems block merge: Test (Python 3.10) failed, and termstory/reminder.py is an unrelated PID-file change bundled in the same PR. Drop the reminder edit and get CI green.

bitflicker64 (changes requested):
Summary: the parser half is right. `_assign_missing_timestamps_fallback` now uses `get_current_time().timestamp()` instead of `time.time()`, and `tests/test_parser.py` adds `test_bash_history_honors_date_override_not_wall_clock`. That is what #450 asked for.

This PR also edits `termstory/reminder.py` and `tests/test_reminder.py` (empty PID-file grace). That is unrelated to #450 and collides with #497 / #500. Drop the reminder files from this branch and keep the parser change only.

bitflicker64 (approved):
Summary: both wall-clock `time.time()` calls in `_assign_missing_timestamps_fallback` now use `get_current_time().timestamp()`. That is what #450 asked for. The new test pins `TERMSTORY_DATE_OVERRIDE` to 2027 while forcing `time.time` to a 2023 epoch, so bash history timestamps stay in range instead of being treated as future.

Earlier reminder.py edits are gone. Diff is parser.py plus one regression in tests/test_parser.py. No unrelated files.

Does what #450 asked. CI: Lint pass, Test 3.9/3.10/3.11/3.12 pass. Mergeable against main.

PR #490, opened 2026-08-28
[Parser] Replace invalid UTF-8 in fish/powershell history

### LINKED ISSUE #451: fix(parser): fish and powershell history drop non-utf8 bytes via errors=ignore
## Context

`termstory/parser.py` `parse_zsh_history` opens history with `encoding='utf-8', errors='replace'` and documents why: Zsh metafied bytes must become a safe placeholder instead of being eaten. `parse_fish_history` and `parse_powershell_history` still open with `errors='ignore'`.

## Problem

Any non-UTF-8 byte sequence in fish or PowerShell history is silently discarded. That can corrupt command text, merge adjacent lines incorrectly, or drop entire commands. The zsh path already fixed this class of bug.

## Expected

Fish and PowerShell parsers should use `errors='replace'` (same as zsh) so invalid bytes become U+FFFD placeholders and the rest of the line is preserved.

## Acceptance criteria

- [ ] Both open calls in `parse_fish_history` and `parse_powershell_history` use `errors='replace'`
- [ ] Test in `tests/test_parser.py` feeds a fish (or powershell) fixture containing a non-UTF-8 byte and asserts the command is still returned with a replacement character rather than being dropped

## Scope

1 file plus a small test; roughly 5-15 lines.

sanjana2505006:
## Summary
* Open fish and PowerShell history with `errors='replace'` (same as zsh) instead of `errors='ignore'`
* Keep non-UTF-8 bytes as U+FFFD so command text is not silently eaten
* Add a fish regression test with an invalid byte sequence

Fixes #451

## Test plan
* [ ] Confirm `test_parse_fish_history_replaces_invalid_utf8` passes

### DIFF
--- a/termstory/parser.py
+++ b/termstory/parser.py
@@ -774,7 +774,7 @@ def parse_fish_history(
         mtime = int(get_current_time().timestamp())
         
     temp_commands = []
-    with open(filepath, 'r', encoding='utf-8', errors='ignore') as f:
+    with open(filepath, 'r', encoding='utf-8', errors='replace') as f:
         current_cmd = None
         current_when = None
         
@@ -824,7 +824,7 @@ def parse_powershell_history(
         mtime = int(get_current_time().timestamp())
         
     raw_lines = []
-    with open(filepath, 'r', encoding='utf-8', errors='ignore') as f:
+    with open(filepath, 'r', encoding='utf-8', errors='replace') as f:
         for line in f:
             raw_lines.append(line)
             
--- a/termstory/reminder.py
+++ b/termstory/reminder.py
@@ -563,7 +563,11 @@ def start_sleep_daemon(db_path: str):
             # is (or is about to become) the running daemon — defer to it.
             try:
                 with open(pid_file, "r") as f:
-                    pid = int(f.read().strip())
+                    content = f.read().strip()
+                if not content:
+                    # Winner created the file but has not written its PID yet.
+                    return
+                pid = int(content)
                 os.kill(pid, 0)
                 return  # Already running
             except (ValueError, OSError):
--- a/termstory/reminder.py
+++ b/termstory/reminder.py
@@ -563,11 +563,7 @@ def start_sleep_daemon(db_path: str):
             # is (or is about to become) the running daemon — defer to it.
             try:
                 with open(pid_file, "r") as f:
-                    content = f.read().strip()
-                if not content:
-                    # Winner created the file but has not written its PID yet.
-                    return
-                pid = int(content)
+                    pid = int(f.read().strip())
                 os.kill(pid, 0)
                 return  # Already running
             except (ValueError, OSError):
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -290,6 +290,16 @@ def test_parse_powershell_history(tmp_path):
     assert commands[0].command == "git status"
     assert commands[1].command == "docker ps ` -a"
 
+
+def test_parse_powershell_history_replaces_invalid_utf8(tmp_path):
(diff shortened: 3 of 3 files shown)


sanjana2505006:
:miku /ecsoc

bitflicker64 (commented):
The change itself is right, but CI is red, so I am holding the merge.

## The diff

`termstory/parser.py:777` and `termstory/parser.py:827` swap `errors='ignore'` for `errors='replace'` on the fish and PowerShell readers. That matches what #451 asked for and it matches the zsh reader, so the three paths now agree. `errors='ignore'` deleting the bad bytes is the actual bug: `echo hello\xffworld` silently became `echo helloworld`, which is a different command from the one the user ran, and nothing downstream could tell that anything had been lost. U+FFFD at least preserves the length and makes the damage visible.

`tests/test_parser.py:268` and `tests/test_parser.py:295` both use `write_bytes` with a real `\xff`, which is the only way to test this. Writing text would go through the encoder and never produce the invalid sequence. Covering both parsers rather than just fish is the right call given both lines changed.

Two files, two lines of production change, no behaviour touched beyond decoding. No objection to any of it.

## Why this is not merged

`Test (Python 3.12)` failed on the latest run:

```
FAILED tests/test_reminder.py::test_start_sleep_daemon_concurrent_invocations_spawn_once - assert 3 == 1
```

That is unrelated to this diff. It is the sleep daemon PID file race, and the traceback is the `FileExistsError` then `ValueError: invalid literal for int() with base 10: ''` sequence in `termstory/reminder.py:560-576`, where a concurrent invocation reads the PID file in the window after `os.open` and before the PID is written into it. It is flaky rather than deterministic: the same test failed on 3.9 and 3.10 on another PR in this batch while passing on 3.12 there, and it passes on 3.12 on several others. Related to #427.

My rule for this repo is that required checks have to be green on the commit being merged, and I will not make an exception for a flake, since the whole point of the rule is that I do not get to decide which red is safe. I also cannot run the suite locally to confirm, so CI is the only test signal I have.

Push an empty commit or ask me to re-run the job and it should go green. Once it does this is ready to merge as-is, no changes needed from you.

sanjana2505006:
@bitflicker64 re-run

bitflicker64 (changes requested):
The parser UTF-8 change looks correct and matches the zsh path. The regression tests cover the intended case.

However the diff also includes an unrelated change in termstory/reminder.py around PID file handling. That is outside the scope of #451 and the PR title. Please remove the reminder.py edit so this PR stays focused on the fish/powershell encoding fix.

bitflicker64 (changes requested):
The #451 change is correct. termstory/parser.py fish and PowerShell readers now use errors='replace', matching zsh. tests/test_parser.py writes raw 0xff bytes and asserts U+FFFD.

termstory/reminder.py still changes empty PID-file handling. That is #427 territory, not this PR. Remove it.

termstory/parser.py
```diff
@@ -774,7 +774,7 @@ def parse_fish_history(
         mtime = int(get_current_time().timestamp())
         
     temp_commands = []
-    with open(filepath, 'r', encoding='utf-8', errors='ignore') as f:
+    with open(filepath, 'r', encoding='utf-8', errors='replace') as f:
```
bitflicker64, line 777: errors='replace' matches the zsh reader and keeps a U+FFFD marker instead of dropping the byte. Fish and PowerShell now behave the same.

bitflicker64 (approved):
Matches #451. Fish and PowerShell history now use errors='replace' like zsh, and the tests check U+FFFD is kept. Small, focused, CI green.

PR #491, opened 2026-08-28
[Parser] Run TimestampDetective on bash/fish/powershell history

### LINKED ISSUE #453: feat(parser): run TimestampDetective on bash, fish, and powershell history
## Context

`termstory/parser.py` `parse_zsh_history` runs `TimestampDetective.resolve_all` on legacy (untimestamped) items so git log, file mtimes, package managers, docker, and lockfiles can recover real dates. `parse_bash_history`, `parse_fish_history`, and `parse_powershell_history` only call `_assign_missing_timestamps_fallback`, which spreads synthetic timestamps from file mtime with no forensic recovery.

## Problem

Users whose primary shell is bash, fish, or PowerShell never get the forensic recovery that zsh users get. Missing timestamps stay synthetic even when git commits or package installs would have provided real anchors.

## Expected

When a bash/fish/powershell history has missing timestamps, pass those items through `TimestampDetective` the same way the zsh path does, then fall back to the existing step-back/interpolation only for items the detective cannot place.

## Acceptance criteria

- [ ] Bash (and ideally fish/powershell) path invokes `TimestampDetective` for untimestamped items
- [ ] Detective-resolved items get `is_legacy=False` and a `recovery_source`
- [ ] Regression coverage in `tests/test_parser.py` or `tests/test_timestamp_detective.py` with a bash fixture that has a recoverable git/file anchor

## Scope

Primarily `termstory/parser.py` plus tests; roughly 40-80 lines if reusing the zsh structure.

sanjana2505006:
## Summary
* Run `TimestampDetective` on untimestamped bash/fish/powershell commands before the synthetic fallback
* Detective-resolved items get `is_legacy=False` and a `recovery_source` (zsh parity)
* Items the detective cannot place still use `_assign_missing_timestamps_fallback`
* Add a bash regression test with a recoverable npm lockfile anchor

Fixes #453

## Test plan
* [ ] Confirm `test_bash_history_timestamp_detective_recovers_file_anchor` passes

### DIFF
--- a/termstory/parser.py
+++ b/termstory/parser.py
@@ -599,10 +599,12 @@ def _assign_missing_timestamps_with_detective(
 
     detective_commands: List[Command] = []
     claimed: set = set()
+    # Shared across detective + fallback so repeated commands don't both take
+    # existing_lookup[cmd][0] and later collapse under exact deduplication.
+    consumed: Dict[str, int] = {}
 
     if legacy_items:
         enriched = _run_timestamp_detective(legacy_items, project_paths)
-        consumed: Dict[str, int] = {}
 
         def resolve_timestamp(cmd: str, fallback_ts: int) -> int:
             if existing_lookup and cmd in existing_lookup:
@@ -632,7 +634,9 @@ def _assign_missing_timestamps_with_detective(
 
     remaining = [(t, cmd) for i, (t, cmd) in enumerate(temp_commands) if i not in claimed]
     fallback_commands = (
-        _assign_missing_timestamps_fallback(remaining, mtime, existing_lookup)
+        _assign_missing_timestamps_fallback(
+            remaining, mtime, existing_lookup, consumed=consumed
+        )
         if remaining else []
     )
 
@@ -652,7 +656,8 @@ def _assign_missing_timestamps_with_detective(
 def _assign_missing_timestamps_fallback(
     temp_commands: List[tuple],
     mtime: int,
-    existing_lookup: Optional[Dict[str, List[int]]]
+    existing_lookup: Optional[Dict[str, List[int]]],
+    consumed: Optional[Dict[str, int]] = None,
 ) -> List[Command]:
     commands_to_return = []
     
@@ -673,7 +678,8 @@ def _assign_missing_timestamps_fallback(
     
     has_any_timestamps = any(t is not None for t, _ in temp_commands)
     
-    consumed = {}
+    if consumed is None:
+        consumed = {}
     def resolve_timestamp(cmd: str, fallback_ts: int) -> int:
         if existing_lookup and cmd in existing_lookup:
             ts_list = existing_lookup[cmd]
--- a/termstory/parser.py
+++ b/termstory/parser.py
@@ -551,7 +551,103 @@ def parse_bash_history(
             if cmd_cleaned:
                 temp_commands.append((None, cmd_cleaned))
                 
-    return _assign_missing_timestamps_fallback(temp_commands, mtime, existing_lookup)
+    return _assign_missing_timestamps_with_detective(
+        temp_commands, mtime, existing_lookup, project_paths
+    )
+
+
+def _run_timestamp_detective(
+    legacy_items: List[dict],
+    project_paths: Optional[Union[List[str], Callable[[], List[str]]]] = None,
+) -> List[dict]:
+    """Run TimestampDetective on untimestamped items (shared by bash/fish/powershell)."""
+    resolved_paths: List[str] = []
+    if project_paths:
+        try:
+            resolved_paths = project_paths() if callable(project_paths) else list(project_paths)
+        except Exception:
+            logger.exception(
+                "TimestampDetective: project_paths callback raised; "
+                "continuing without project-path enrichment."
+            )
+    from termstory.config import load_config
+    _td_config = load_config()
+    detective = TimestampDetective(
+        search_root=os.path.expanduser("~"),
+        project_paths=resolved_paths or [],
+        macos_install_log=_td_config.get("macos_install_log", "/private/var/log/install.log"),
+    )
+    return detective.resolve_all(legacy_items)
+
+
+def _assign_missing_timestamps_with_detective(
+    temp_commands: List[tuple],
+    mtime: int,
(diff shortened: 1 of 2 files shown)


sanjana2505006:
:miku /ecsoc

bitflicker64 (commented):
This is a careful port and I could not find a correctness bug in it. I am still not merging it on one pass, because of what it does to the bash, fish, and PowerShell parse paths rather than because of how it is written.

## What I checked and liked

The parity with the zsh path is genuinely close, and I went through it line by line rather than taking the description's word for it.

The tail of `_assign_missing_timestamps_with_detective` at `termstory/parser.py:648-661` is the same `five_years_ago <= ts <= now` filter and `sort(key=lambda x: x.timestamp)` that `parse_zsh_history` already applies at `termstory/parser.py:448-459`. Copying that rather than inventing new bounds is the right instinct, and it is why I am not worried about the new filter dropping things the zsh path would have kept.

The `is_legacy` handling is right in the way that is easy to get wrong. Rather than keeping only items the detective fully resolved, `termstory/parser.py:619` takes everything with a non-null `detected_ts` and sets `is_legacy = not is_real`. That matches zsh, which splits `still_synthetic` into `interpolated` and `unresolvable` at `termstory/parser.py:456-457` and still emits the interpolated ones with `is_legacy=True`. Dropping them instead would have quietly lost commands the detective had placed between two anchors.

Threading `consumed` through into `_assign_missing_timestamps_fallback` at `termstory/parser.py:635` is the subtle fix in this PR and it is correct. Without it the detective pass and the fallback pass each start their own counter, both hand the same repeated command `existing_lookup[cmd][0]`, and the duplicates collapse under exact deduplication downstream. `test_fish_history_shares_timestamp_locks_across_detective_and_fallback` pins exactly that, and the mock detective is the right tool since the real one cannot be made to resolve deterministically in a test.

`duration=None` at `termstory/parser.py:617` differs from zsh's `duration=0`, but it matches what the bash fallback already emits at both of its `Command` construction sites, so it is consistent with the path it is on. No objection.

The `project_paths` callback being wrapped in a broad `try/except` with a logged exception mirrors the zsh handling and the reasoning in the comment there. Fine.

## Why it is escalated rather than merged

Before this change, parsing a bash, fish, or PowerShell history was filesystem reads and regex, nothing more. After it, `_run_timestamp_detective` at `termstory/parser.py:562` puts `TimestampDetective.resolve_all` on all three of those paths, and the detective shells out: `git log`, docker inspection, Homebrew Cellar probing, and on macOS it reads `/private/var/log/install.log`. `termstory/timestamp_detective.py` has half a dozen `subprocess.run` sites.

Anything that causes external processes to run gets a manual review in this repo before it lands. It does not matter that the diff adds no `subprocess` import of its own and reuses a module already wired to zsh. What matters is that three parsers that never spawned a process now will, and that is the kind of change I want to be deliberate about rather than approve because the checks are green.

Two specific things I want to look at before taking it:

1. Cost on the common case. `search_root` is `os.path.expanduser("~")` and a default bash history with no `HISTTIMEFORMAT` means every single command arrives untimestamped, so `legacy_items` is the entire file. For a long history that is the detective's full detector suite over the whole home directory, on a code path that was previously cheap. The zsh path pays this too, but zsh histories usually carry real timestamps, so it is the exception there and it would be the rule here. I want to see a timing on a realistic multi-thousand-line `.bash_history` before I sign off.

2. `os.environ["TERMSTORY_MISSING_TIMESTAMPS"] = "1"` at `termstory/parser.py:591`. This mirrors `termstory/parser.py:238` on the zsh side, so it is parity, but it is a process-global mutation that is now set from three more entry points and is never cleared. Worth confirming nothing reads it in a way that gets confused when several parsers run in one process.

Neither is a defect in your diff. Both are consequences of extending this machinery to paths that did not have it, and they are my call to make, not yours.

## CI

Green on 3.9 through 3.12 plus lint. I have not run the suite locally, so that is the only test signal I have.

Nothing for you to change right now. I will come back to this with the timing numbers.

bitflicker64 (commented):
This extends TimestampDetective usage to bash/fish/powershell and adds shared helper logic plus regression tests. The approach mirrors the existing zsh path.

Because TimestampDetective performs multiple subprocess calls (git, brew, npm, docker, last) this PR introduces new subprocess execution paths. It needs a manual look before any merge decision.

bitflicker64 (commented):
Matches #453: bash/fish/powershell now go through _assign_missing_timestamps_with_detective before the synthetic fallback. consumed is shared with the fallback so repeated commands do not steal the same existing_lookup slot. Tests in tests/test_parser.py pin the npm lockfile recover path and the shared-lock case.

Not merging. _run_timestamp_detective in termstory/parser.py puts TimestampDetective.resolve_all on three parsers that previously only read files. That detective shells out (git, brew, npm, docker, last). Needs a manual look.

bitflicker64:
Holding this for a manual look. TimestampDetective on bash/fish/powershell adds subprocess execution to parsers that previously only read history files.

termstory/parser.py
```diff
@@ -551,12 +551,113 @@ def parse_bash_history(
...
+    """Resolve missing timestamps via TimestampDetective, then fall back for the rest."""
+    legacy_items = []
+    legacy_indices = []
+    for i, (t, cmd) in enumerate(temp_commands):
+        if t is None:
+            legacy_items.append({"timestamp": None, "duration": None, "command": cmd})
+            legacy_indices.append(i)
+
+    if legacy_items:
+        os.environ["TERMSTORY_MISSING_TIMESTAMPS"] = "1"
```
bitflicker64, line 598: This writes TERMSTORY_MISSING_TIMESTAMPS into the process environment as a side effect of parsing. That matches the zsh path, but it should stay scoped to the parse call rather than leaking into later code in the same process.

bitflicker64 (approved):
Does what #453 asked: bash, fish, and powershell now run TimestampDetective before the synthetic fallback, with is_legacy/recovery_source on detective hits. CI is green. The TERMSTORY_MISSING_TIMESTAMPS env write is inherited from the zsh path; worth keeping an eye on, not a merge blocker.

PR #496, opened 2026-08-29
Fix/494 deterministic command ordering

### LINKED ISSUE #494: Timeline/replay ordering is nondeterministic for equal timestamps
## Context

Termstory stores timestamps for commands and later uses those records in timeline, replay, and related presentation paths.

## Problem

Multiple commands can legitimately have the same timestamp, particularly when commands are recorded with coarse timestamp resolution or are persisted within the same timestamp boundary.

If a query orders records only by timestamp, records with identical timestamps have no explicit secondary ordering guarantee. SQLite does not guarantee a stable order for rows whose ORDER BY keys are equal.

For example:

command A → timestamp 1000
command B → timestamp 1000
command C → timestamp 1001

The expected order should remain A, B, C whenever the same database is queried repeatedly, rather than allowing A/B to depend on incidental database row ordering.

## Expected behavior

Commands with identical timestamps should have a deterministic secondary ordering based on an appropriate persisted identity/order field.

The ordering should remain stable across:

- repeated database queries
- fresh database connections
- timeline output
- replay
- other consumers that rely on the same command ordering

Commands with different timestamps must continue to follow chronological ordering.

## Scope

First inspect the existing database queries and all relevant timeline/replay consumers to determine whether a deterministic secondary ordering is already guaranteed.

If it is missing, introduce the smallest appropriate ordering change and regression tests.

Do not alter timestamp generation, session recovery, export date-boundary semantics, or unrelated database behavior.

If this aligns with the project's direction, please assign this issue to me.

deepmhatre13:
Closes #494
# Fix #494: Make Equal-Timestamp Command Ordering Deterministic

## Summary

Fix nondeterministic command ordering when multiple persisted commands have the same timestamp.

SQLite does not guarantee ordering between rows whose `timestamp` values are equal unless an explicit secondary ordering key is provided. This could cause timeline/replay consumers to receive commands in an inconsistent order.

This PR uses the persisted `commands.id` as the deterministic secondary ordering key while preserving chronological ordering for commands with different timestamps.

## Root Cause

The canonical command-fetching query in `termstory/database.py` ordered commands only by:

```sql
ORDER BY timestamp ASC
```

When multiple commands had identical timestamps, their relative order was left to SQLite's unspecified row ordering.

Additionally, `termstory/replay.py` sorted commands using only:

```python
sorted(session.commands, key=lambda c: c.timestamp)
```

Python's stable sort preserved whatever ordering had already been produced by the database, meaning an unstable database ordering could propagate directly into replay.

### Database command ordering

Updated the canonical command retrieval query in `termstory/database.py` from:

```sql
ORDER BY timestamp ASC
```

to:

```sql
ORDER BY timestamp ASC, id ASC
```

This establishes:

1. `timestamp ASC` — chronological ordering remains the primary criterion.
2. `id ASC` — persisted command identity deterministically resolves equal timestamps.

### Replay ordering

Updated `run_replay()` in `termstory/replay.py` from timestamp-only sorting to:

```python
sorted(session.commands, key=lambda c: (c.timestamp, c.id))
```

This ensures replay remains deterministic even when commands share the same timestamp.

### Timeline behavior

No direct change was required in `termstory/timeline.py`.

Timeline consumers obtain commands through the database session-building path, which now provides deterministic `(timestamp, id)` ordering. Timeline rendering itself does not independently reorder commands.

### Parser behavior

`termstory/parser.py` was intentionally left unchanged.

Parser-level timestamp sorting occurs before commands are persisted and therefore before their database IDs are assigned. Using `commands.id` at that stage would not provide a meaningful persisted ordering key.

## Regression Tests

Added coverage for:

* Two commands with identical timestamps being ordered by `id`
* Commands with different timestamps remaining chronologically ordered
* Deterministic `session.commands` ordering across fresh database connections
* Persisted command IDs being populated and distinct
* Replay ordering equal-timestamp commands by ID
* Replay preserving chronological ordering when timestamps differ

The replay regression test was also verified against the old implementation: reverting the replay change causes the equal-timestamp regression test to fail, confirming that the test catches the original bug.

## Validation

Focused tests:

```text
tests/test_replay.py       11 passed
tests/test_timeline.py      4 passed
tests/test_database.py     29 passed
```

Broader relevant test suite:

```text
214 passed
```

No existing tests were weakened or removed.

## Scope

No changes were made to:

* Timestamp generation
* Session recovery
* Export date-boundary logic
* Backup/restore behavior
* MCP snapshot handling
* Search ranking / issue #493
* Deduplication or migration logic
* Parser pre-persistence ordering
* Database schema

No migration is required because `commands.id` already exists as the persisted primary key.

### DIFF
--- a/termstory/database.py
+++ b/termstory/database.py
@@ -858,7 +858,7 @@ class Database:
             SELECT id, timestamp, command, exit_code, session_id, project_id, recovery_source, is_legacy
             FROM commands
             WHERE session_id IN ({s_ids_str})
-            ORDER BY timestamp ASC
+            ORDER BY timestamp ASC, id ASC
         """, session_ids)
         all_cmd_rows = cursor.fetchall()
         
--- a/termstory/replay.py
+++ b/termstory/replay.py
@@ -106,7 +106,7 @@ def run_replay(db: Database, session_id: Optional[int] = None, speed: float = 1.
         if projects:
             project_name = projects[0].name
 
-    commands = sorted(session.commands, key=lambda c: c.timestamp)
+    commands = sorted(session.commands, key=lambda c: (c.timestamp, c.id or 0))
     if not commands:
         console.print(f"[yellow]Session #{session_id} has no commands to replay.[/yellow]")
         return
--- a/tests/test_database.py
+++ b/tests/test_database.py
@@ -930,3 +930,163 @@ def test_recovery_preserves_command_and_session_identity(tmp_path):
 
     # Session→command relationships are intact.
     assert {r[3] for r in commands_after} == {1, 2}
+
+
+# ---------------------------------------------------------------------------
+# Issue #494: deterministic command ordering for equal timestamps (id ASC)
+# ---------------------------------------------------------------------------
+
+def _seed_session_for_ordering(db, session_id, start_ts, command_pairs):
+    """Insert a session row and commands with explicit (timestamp, command) values.
+
+    ``command_pairs`` is a list of (timestamp, command_text) tuples. Commands are
+    inserted in the given order so their auto-incremented ids reflect that order.
+    Returns a dict mapping command text -> persisted id.
+    """
+    conn = db.get_connection()
+    try:
+        cursor = conn.cursor()
+        cursor.execute(
+            "INSERT INTO sessions (id, start_time, end_time, duration_seconds) "
+            "VALUES (?, ?, ?, ?)",
+            (session_id, start_ts, start_ts, 0),
+        )
+        id_by_cmd = {}
+        for ts, cmd_text in command_pairs:
+            cursor.execute(
+                "INSERT INTO commands (timestamp, command, exit_code, session_id, project_id, is_legacy) "
+                "VALUES (?, ?, ?, ?, ?, ?)",
+                (ts, cmd_text, 0, session_id, None, 0),
+            )
+            id_by_cmd[cmd_text] = cursor.lastrowid
+        conn.commit()
+        return id_by_cmd
+    finally:
+        conn.close()
+
+
+def test_command_ordering_tie_break_by_id(tmp_path):
+    """Equal timestamps must be ordered by persisted id ASC (not rowid/text)."""
+    db_file = tmp_path / "test_ordering_tie.db"
+    db = Database(str(db_file))
+    db.init_db()
+
+    # Insert two commands with identical timestamps but deliberately reversed
+    # command-text ordering relative to insertion order, so the only stable
+    # tiebreaker is the persisted id (insertion order).
+    session_id = 1
+    id_by_cmd = _seed_session_for_ordering(
+        db, session_id=session_id, start_ts=1000,
+        command_pairs=[
+            (1000, "cmd-first-id-lower"),
+            (1000, "cmd-second-id-higher"),
+        ],
+    )
+
+    # The lower id exists and precedes the higher id.
+    lower_id = id_by_cmd["cmd-first-id-lower"]
+    higher_id = id_by_cmd["cmd-second-id-higher"]
+    assert lower_id < higher_id
+
+    # Retrieve through the production session-building path (get_sessions_by_ids
+    # -> _build_sessions) with fresh connections: ordering must be stable and
+    # id-based for equal timestamps.
+    for _ in range(50):
+        sessions = Database(str(db_file)).get_sessions_by_ids([session_id])
+        assert len(sessions) == 1
+        cmds = sessions[0].commands
+        assert len(cmds) == 2
+        # lower id must appear first despite identical timestamps.
+        assert cmds[0].id == lower_id
+        assert cmds[1].id == higher_id
+        assert cmds[0].command == "cmd-first-id-lower"
+        assert cmds[1].command == "cmd-second-id-higher"
+
+
+def test_command_ordering_different_timestamps_not_overridden_by_id(tmp_path):
+    """When timestamps differ, chronological ordering wins over id ordering.
+
+    Scenario from the issue:
+        id 1 -> ts 1000
+        id 2 -> ts 1000
+        id 3 -> ts 1001
+    Expected traversal order: 1, 2, 3.
+    """
+    db_file = tmp_path / "test_ordering_chrono.db"
+    db = Database(str(db_file))
+    db.init_db()
+
+    session_id = 1
+    id_by_cmd = _seed_session_for_ordering(
+        db, session_id=session_id, start_ts=1000,
+        command_pairs=[
+            (1000, "cmd-ts-1000-a"),   # id 1
+            (1000, "cmd-ts-1000-b"),   # id 2
+            (1001, "cmd-ts-1001-c"),   # id 3
+        ],
+    )
+    expected_id_order = [
+        id_by_cmd["cmd-ts-1000-a"],
+        id_by_cmd["cmd-ts-1000-b"],
+        id_by_cmd["cmd-ts-1001-c"],
+    ]
+
+    for _ in range(50):
+        sessions = Database(str(db_file)).get_sessions_by_ids([session_id])
+        assert len(sessions) == 1
+        actual_id_order = [c.id for c in sessions[0].commands]
+        assert actual_id_order == expected_id_order
+
+
+def test_session_commands_deterministic_across_fresh_connections(tmp_path):
+    """session.commands (via get_sessions_by_ids -> _build_sessions) must be
+    deterministically ordered (timestamp ASC, id ASC) even when re-queried on
+    a brand-new database connection."""
+    db_file = tmp_path / "test_session_ordering.db"
+    db = Database(str(db_file))
+    db.init_db()
+
+    _seed_session_for_ordering(
+        db, session_id=1, start_ts=1000,
+        command_pairs=[
+            (1000, "cmd-a"),
+            (1000, "cmd-b"),
+            (1000, "cmd-c"),
+            (1001, "cmd-d"),
+        ],
+    )
+
+    # Expected: equal-timestamp group ordered by id ASC, then the ts=1001 cmd.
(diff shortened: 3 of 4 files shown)


deepmhatre13:
:miku /ecsoc

deepmhatre13:
@bitflicker64 ,kindly review the pr

bitflicker64 (changes requested):
The core ordering fix (ORDER BY timestamp ASC, id ASC in database.py and the matching sort in replay.py) correctly addresses #494 and the new tests exercise the equal-timestamp case.

However the same diff also lands the full #493 exactness-tier ranking implementation in database.py and search.py plus a large block of #493 tests. That is unrelated to the PR title and linked issue. Please split the #493 work into its own PR so this one stays limited to deterministic command ordering.

bitflicker64 (changes requested):
Still blocked on scope.

The #494 bits are correct: termstory/database.py ORDER BY timestamp ASC, id ASC and termstory/replay.py sorting on (c.timestamp, c.id).

The same diff still ships _exactness_tier plus the search ranking wiring and tests/test_search.py coverage for #493. That is a different issue. Split it out.

deepmhatre13:
Cleaned up PR #496 to contain only the #494 deterministic command-ordering changes.

Removed the unrelated #493 commits from the PR history and rebuilt the branch with only the #494 implementation and regression tests.

Validation:

* Database + replay + timeline tests: **44 passed**
* `git diff --check`: clean
* PR diff is limited to `database.py`, `replay.py`, `test_database.py`, and `test_replay.py`.

The branch was force-updated with `--force-with-lease`.

deepmhatre13:
Investigated the remaining Python 3.10 CI failure.

The failing test is `tests/test_reminder.py::test_start_sleep_daemon_concurrent_invocations_spawn_once`, which is unrelated to #494.

I reproduced the test locally on the #494 branch and it passes. I also checked the same test against the clean #494 base commit (`9248676`) in a separate worktree, where it also passes.

The #494-focused changes remain isolated to the deterministic command-ordering implementation and regression tests, with the relevant database/replay/timeline tests passing (44 passed).

I have intentionally not modified the reminder code or tests, since doing so would introduce an unrelated change into #494.

termstory/database.py
```diff
@@ -858,7 +858,7 @@ def _build_sessions(self, cursor, session_rows) -> List[Session]:
             SELECT id, timestamp, command, exit_code, session_id, project_id, recovery_source, is_legacy
             FROM commands
             WHERE session_id IN ({s_ids_str})
-            ORDER BY timestamp ASC
+            ORDER BY timestamp ASC, id ASC
```
bitflicker64, line 861: ORDER BY timestamp ASC, id ASC is the right tie-break for _build_sessions. CI is not green: Test (Python 3.10) failed. Also tests/test_database.py has a stray indent change around the existing recovery assertion that needs to be checked.

bitflicker64 (changes requested):
The (timestamp, id) ordering in database.py and replay.py is the right fix for #494. Cannot merge: Test (Python 3.10) failed and mergeable_state is unstable. Also check the indent change in tests/test_database.py around the existing recovery assertion.

deepmhatre13 (commented):
.

bitflicker64 (changes requested):
Summary: `ORDER BY timestamp ASC, id ASC` in `termstory/database.py` `_build_sessions` and `sorted(..., key=lambda c: (c.timestamp, c.id))` in `termstory/replay.py` match #494.

Problems:
- Required check Test (Python 3.10) failed, so this is not mergeable.
- Replay sort will raise if `c.id` is None (`(timestamp, None)` is not ordered against an int in Python 3). Guard that key, e.g. `(c.timestamp, c.id if c.id is not None else 0)`.
- `tests/test_database.py` only changed the indent on the "Session relationships are intact" comment next to an existing assert. Keep that file clean.

Does the linked issue work: yes for persisted rows with ids. Needs the None-id guard and a green 3.10 run.

bitflicker64 (changes requested):
Summary: the production change is the right idea for #494. `_build_sessions` now orders `timestamp ASC, id ASC`, and `run_replay` sorts on `(c.timestamp, c.id)`.

Two blockers:

1. Test (Python 3.10) failed. Required checks are not all green.
2. `run_replay` will raise TypeError if any command has `id is None`. Python 3 cannot compare None with int inside the sort key. Parser-built or unsaved Command objects can have a missing id. Guard that case (treat missing id as a last-resort tiebreaker) or document that replay only runs on persisted rows with ids.

The new database tests mostly re-query with the same ORDER BY they are asserting, so they do not fully exercise `_build_sessions`. The session-level tests that go through `get_sessions_by_ids` are the useful ones.

CI: Lint pass, Test 3.9/3.11/3.12 pass, Test 3.10 fail. mergeable_state=unknown.

termstory/replay.py
```diff
@@ -106,7 +106,7 @@ def run_replay(db: Database, session_id: Optional[int] = None, speed: float = 1.
         if projects:
             project_name = projects[0].name
 
-    commands = sorted(session.commands, key=lambda c: c.timestamp)
+    commands = sorted(session.commands, key=lambda c: (c.timestamp, c.id))
```
bitflicker64: `sorted(..., key=lambda c: (c.timestamp, c.id))` raises TypeError in Python 3 when `c.id` is None. Replay can see unsaved Command objects. Treat a missing id as a last-resort tiebreaker, or drop the in-memory sort and keep the DB order from `_build_sessions`.

tests/test_database.py
```diff
@@ -928,5 +928,181 @@ def test_recovery_preserves_command_and_session_identity(tmp_path):
...
+    lower_id = id_by_cmd["cmd-first-id-lower"]
+    higher_id = id_by_cmd["cmd-second-id-higher"]
+    assert lower_id < higher_id
+
+    # Query repeatedly with fresh connections: ordering must be stable and id-based.
+    for _ in range(50):
+        conn = db.get_connection()
+        try:
+            cursor = conn.cursor()
+            cursor.execute(
```
bitflicker64: This query repeats the same ORDER BY the production change added. It does not call `_build_sessions` or `get_sessions_by_ids`, so it would still pass if the production query were left unchanged. Prefer asserting through `get_sessions_by_ids` like the later tests do.

bitflicker64 (changes requested):
Inline notes on the None id sort key and the tests that restate the ORDER BY instead of going through `_build_sessions`.

termstory/database.py
bitflicker64, line 861: This is the right fix for #494. `ORDER BY timestamp ASC, id ASC` is the deterministic secondary key the issue asked for.

termstory/replay.py
bitflicker64: `c.id` can be None on in-memory Command objects that never went through the DB. `sorted(..., key=lambda c: (c.timestamp, c.id or 0))` avoids a TypeError on mixed None/int keys in Python 3.

bitflicker64 (commented):
Summary: database fetch now uses `ORDER BY timestamp ASC, id ASC`, replay sorts on `(timestamp, id)`. That is what #494 asked for. Timeline correctly inherits the DB order. Tests cover equal-ts and chrono cases.

Guard `c.id` when it is None. CI: Test (Python 3.10) failed; other versions and Lint passed.

Not merging until 3.10 is green.

bitflicker64 (changes requested):
Summary: ORDER BY timestamp ASC, id ASC in database.py and the matching (timestamp, id) sort in replay.py do what #494 asked. Tests cover the tie-break path.

CI: Test (Python 3.10) is failing. Required checks must all pass before this can land.

bitflicker64 (changes requested):
Summary: database._build_sessions now ORDER BY timestamp ASC, id ASC. replay.run_replay sorts by (timestamp, id). That is what #494 asked for. Scope is the two production lines plus tests.

termstory/replay.py: sorted(..., key=lambda c: (c.timestamp, c.id)) will raise if c.id is None. Guard that, or keep timestamp-only when id is missing.

tests/test_database.py test_command_ordering_tie_break_by_id: the 50-iteration loop re-runs the ORDER BY in raw SQL. It does not go through _build_sessions. test_session_commands_deterministic_across_fresh_connections does, so keep that one and drop the raw-SQL loops or point them at get_sessions_by_ids.

tests/test_replay.py is missing a trailing newline.

Does what #494 asked. Cannot merge: Test (Python 3.10) failed. Lint and 3.9/3.11/3.12 passed. mergeable_state=unstable.

bitflicker64 (approved):
Looks correct. The ORDER BY timestamp ASC, id ASC in database.py:861 and the matching (timestamp, id or 0) sort in replay.py:109 fix the nondeterministic equal-timestamp ordering described in #494. Tests cover the tie-break, chrono priority, fresh connections, and None id cases. Diff matches the linked issue with no unrelated changes. CI green.

termstory/database.py
bitflicker64, line 861: This is the right tie-break for #494. `timestamp ASC, id ASC` keeps chronology first and uses the persisted primary key when timestamps collide.

termstory/replay.py
```diff
@@ -106,7 +106,7 @@ def run_replay(db: Database, session_id: Optional[int] = None, speed: float = 1.
         if projects:
             project_name = projects[0].name
 
-    commands = sorted(session.commands, key=lambda c: c.timestamp)
+    commands = sorted(session.commands, key=lambda c: (c.timestamp, c.id or 0))
```
bitflicker64, line 109: `(c.timestamp, c.id or 0)` matches the database order and avoids a TypeError when an in-memory command has `id=None`. Good.

bitflicker64 (commented):
Does what #494 asked. Database fetch now uses `ORDER BY timestamp ASC, id ASC`, replay sorts by `(timestamp, id or 0)`, and the new tests hit the production session path plus a scrambled replay list.

No unrelated files. Holding merge until every required CI job is green on this head.

PR #499, opened 2026-08-29
fix(export): respect per-command project attribution

### LINKED ISSUE #498: fix(export): respect per-command project attribution in project filters
### Description

Termstory supports per-command project attribution for sessions that span multiple projects. A session's `project_id` represents its final project, while individual commands can have their own `project_id`.

The export project filter currently relies only on the session-level `project_id`. As a result, a session that contains commands attributed to the requested project can be incorrectly excluded from the export if the session later moves to another project.

### Reproduction

Consider a session containing:

- Command 1 → attributed to Project A
- Command 2 → attributed to Project A
- Command 3 → attributed to Project B
- Session's final `project_id` → Project B

Running an export filtered to Project A can omit the entire session, even though commands from that session were explicitly attributed to Project A.

### Expected behavior

When exporting with a project filter, a session should be considered a match when:

1. The session's final project matches the requested project, **or**
2. At least one command in the session has `command.project_id` matching the requested project.

This should preserve the existing session-level behavior while also respecting per-command project attribution.

### Scope

- Update the project-filtering logic in the exporter to consider per-command project attribution.
- Preserve existing behavior for sessions that have only a session-level project.
- Do not modify how project attribution is assigned or persisted.
- Add regression tests covering a session that moves between two projects and contains commands attributed to both projects.
- Verify that exporting by the earlier project's name/path includes the session.

### Acceptance Criteria

- A session with a command attributed to Project A is included when exporting with Project A as the filter, even if the session's final project is Project B.
- A session whose session-level and command-level projects do not match the requested project is not incorrectly included.
- Existing export behavior and tests remain unchanged for single-project sessions.
- Regression tests pass on all supported Python versions.

deepmhatre13:
Closes #498 
## Summary

Fixes project-filtered exports to respect per-command project attribution.

Previously, `fetch_export_data()` considered only the session-level `project_id` when applying a project filter. For sessions that moved between projects, this could exclude a session from an earlier project's export even when commands in that session were explicitly attributed to that project.

### Changes

- Project filters now match against:
  - the session-level `project_id`, or
  - any command-level `project_id` belonging to the session.
- Project resolution now includes project IDs referenced by commands.
- Unresolvable command project IDs are ignored rather than incorrectly matching the `"other"` bucket.
- Added regression coverage for session-level, command-level, mixed-project, path, no-project, and false-positive cases.

### Validation

- `tests/test_exporter.py`: **56 passed**
- Search/database/integration tests: **24 passed**
- Database query test: **1 passed**
- Consolidated relevant suite: **80 passed**
- `git diff --check`: clean

No schema or project-attribution persistence changes were made.

### DIFF
--- a/termstory/exporter.py
+++ b/termstory/exporter.py
@@ -146,6 +146,56 @@ def parse_until(until_str: Optional[str]) -> Optional[int]:
         raise ValueError(f"Invalid date or day count format '{until_str}': {e}")
 
 
+def _project_matches_filter(proj: Optional[Project], filter_lower: str) -> bool:
+    """Return True if a resolved :class:`Project` matches the lowercased filter.
+
+    ``proj`` is ``None`` for sessions/commands carrying no project attribution.
+    In that case the filter only matches the conventional "no project" bucket
+    names, preserving the existing behaviour for sessions without a project.
+    """
+    if proj is None:
+        return filter_lower in ("other", "general", "no project")
+    return filter_lower in proj.name.lower() or filter_lower in proj.path.lower()
+
+
+def _session_matches_project_filter(
+    session: Session, project_map: Dict[int, Project], filter_lower: str
+) -> bool:
+    """Return True if *session* matches a project filter.
+
+    A session matches when the requested project is reflected in *either*:
+
+    * the session-level attribution (``session.project_id`` — the final
+      project the session ended in), **or**
+    * the per-command attribution of at least one command belonging to the
+      session (``command.project_id``) — e.g. a session that ``cd``-ed
+      between projects mid-stream.
+
+    This mirrors the per-command project matching already used by
+    :mod:`termstory.search` (see #339) and fixes #498, where the export
+    path only considered the session's final project.
+
+    ``project_map`` maps resolved project IDs to :class:`Project` entities
+    (name/path). Commands whose ``project_id`` is not present in the map are
+    treated as an unresolvable, real project and are *not* matched — this
+    prevents an unattributed/missing command from falsely matching the
+    "other" bucket.
+    """
+    # Session-level attribution (the final/current project).
+    session_proj = project_map.get(session.project_id) if session.project_id is not None else None
+    if _project_matches_filter(session_proj, filter_lower):
+        return True
+    # Per-command attribution: a session matches if any of its commands was
+    # attributed to the requested project, even when the session's final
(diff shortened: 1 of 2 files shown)


deepmhatre13:
:miku /ecsoc

deepmhatre13:
Fixed the Greptile P1 finding.

The export project filter now considers explicit per-command `project_id=None` attribution when resolving the `other` / `general` / `no project` bucket.

This preserves the distinction between:
- an explicitly unattributed command (`project_id is None`) → matches the no-project bucket;
- a command associated with a real project → matches only that project's name/path;
- an unknown/unresolvable project ID → does not incorrectly match the no-project bucket.

Added focused regression coverage for mixed-project sessions, session-level no-project attribution, unknown project IDs, and false-positive text matches.

Validation:
- `tests/test_exporter.py`: 62 passed
- Relevant PR suites: 25 passed
- Combined: 87 passed
- `git diff --check`: clean

The unrelated `test_start_sleep_daemon_concurrent_invocations_spawn_once` CI failure in `tests/test_reminder.py` remains intentionally untouched because it is outside the scope of this export PR.

termstory/exporter.py
```diff
@@ -146,6 +146,68 @@ def parse_until(until_str: Optional[str]) -> Optional[int]:
...
+    unknown/missing project ID from falsely matching the "other" bucket.
+    """
+    # Session-level attribution (the final/current project).
+    session_proj = project_map.get(session.project_id) if session.project_id is not None else None
+    if _project_matches_filter(session_proj, filter_lower):
+        return True
+    # Per-command attribution: a session matches if any of its commands was
+    # attributed to the requested project, even when the session's final
+    # project differs. An explicitly unattributed command (project_id is
+    # None) matches the no-project bucket ("other" / "general" /
```
bitflicker64, line 198: Session-level match still uses the final project_id, then any command-level project_id. That is the #498 gap. Unknown command IDs correctly skip the other bucket because project_map.get returns None and we only match when cmd_proj is not None.

bitflicker64 (approved):
Looks good. fetch_export_data now matches a session when either session.project_id or any command.project_id hits the filter, which is what #498 asked for. Unknown command project IDs do not fall into the other bucket. Tests cover mixed sessions, path matches, and false positives from command text. CI is green (lint + 3.9-3.12).

PR #501, opened 2026-08-30
[Project] Preserve per-command project ids when session ends at home

### LINKED ISSUE #462: fix(project): home cwd at session end wipes per-command project ids
Context

Pass 1 of detect_projects() in termstory/project.py walks each command, resolves the simulated cwd, and writes cmd.project_id from that cwd (see the inline attribution loop and the overwrite_per_command=False call at lines 521-526). That is how a session that cds through more than one project keeps per-command attribution.

When the session's final cwd is home or /, the else branch at termstory/project.py lines 527-530 does this:

    else:
        session.project_id = None
        for cmd in session.commands:
            cmd.project_id = None

Problem

Any session that does real project work and then cds back to ~ before the session gap fires loses every per-command project_id that Pass 1 just computed. Search, stats, and export then treat those commands as unattributed even though the cwd at command time was a real project root.

Reproduction:

1. cd ~/Projects/acme && pytest
2. cd ~ && ls
3. Wait past the session gap and re-ingest.
4. The pytest command is stored with project_id NULL.

Expected

Session.project_id may stay None when the final cwd is home. Commands that ran inside a project root must keep the project_id assigned during the cwd walk. Only commands whose cwd was home/root should stay unattributed.

Acceptance criteria

- [ ] Ending a session at ~ does not clear project_id on earlier commands that ran under a project root.
- [ ] Commands that never left home still have project_id None.
- [ ] Session-level project_id remains the final cwd project (None when that cwd is home).
- [ ] Add a regression test in tests/test_project.py covering cd into a project, a project command, then cd ~.

Scope

One production file (termstory/project.py, ~10 lines) plus one test in tests/test_project.py.

sanjana2505006:
## Summary
- Closes #462
- When final cwd is home/root, keep `session.project_id = None` but stop clearing per-command `project_id` values assigned during Pass 1
- Add regression test: cd into project → run command → cd ~ → ls

## Test plan
- [x] `pytest tests/test_project.py::test_detect_projects_preserves_command_ids_when_session_ends_at_home -q`
- [x] `pytest tests/test_project.py -q`

### DIFF
--- a/termstory/project.py
+++ b/termstory/project.py
@@ -525,9 +525,9 @@ def detect_projects(sessions: List[Session]) -> List[Project]:
                 session, final_project, projects_dict, overwrite_per_command=False
             )
         else:
+            # Final cwd is home/root: session stays unattributed, but per-command
+            # project_ids assigned during the cwd walk must be preserved (#462).
             session.project_id = None
-            for cmd in session.commands:
-                cmd.project_id = None
 
         last_session_end = effective_end
     
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -678,6 +678,40 @@ def test_detect_projects_per_command_null_handling(monkeypatch):
     assert s.commands[2].project_id == real_proj.id
 
 
+def test_detect_projects_preserves_command_ids_when_session_ends_at_home(monkeypatch):
+    """#462: cd back to ~ at session end must not wipe earlier per-command project_ids."""
+    import os
+    original_listdir = os.listdir
+
+    def mock_listdir(path):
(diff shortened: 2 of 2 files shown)


bitflicker64 (approved):
Summary: Pass 1 no longer wipes per-command `project_id` when the final cwd is home/root. `session.project_id` stays None. That is what #462 asked for.

The new test in `tests/test_project.py` covers `cd ~/Projects/acme` -> `pytest` -> `cd ~` -> `ls`. CI is green. Diff stays in `termstory/project.py` and the one regression test.

PR #502, opened 2026-08-30
[Predict] Exclude legacy commands from prediction signals (fixes #463)

### LINKED ISSUE #463: fix(predict): legacy commands leak into prediction signals
Context

termstory/predict.py documents that predictions respect the is_legacy flag and that synthetic timestamps are excluded (module docstring, lines 11-12). Predictor._load_sessions() says it loads "all non-legacy sessions" (lines 78-80).

The sessions query at lines 86-105 has no is_legacy predicate. After fetch, the only filter is at lines 134-135: a session is skipped only when every command in that session is legacy. Surviving sessions then keep every non-noise command, including commands whose is_legacy flag is set:

    cmds = [r[0] for r in cmd_rows if not _is_noise_command(r[0])]

Problem

Mixed sessions (imported history plus later live commands) contribute synthetic-timestamp commands to recency, time-of-day affinity, day cadence, and suggested command lists. That contradicts the module contract and can surface stale imported tools as "likely next" work.

Reproduction:

1. Ingest a session that contains one is_legacy=1 command ("make") and one live command ("pytest").
2. Run termstory predict.
3. The suggested command list can include make even though that timestamp is synthetic.

Expected

Drop any command with is_legacy truthy before scoring. Skip a session when no non-legacy commands remain. Keep the all-legacy session skip as a fast path.

Acceptance criteria

- [ ] _load_sessions() excludes is_legacy commands from the commands list used for scoring.
- [ ] A session that only has legacy commands after that filter is omitted from signals.
- [ ] Recency / affinity / suggested_commands do not include legacy-only tools.
- [ ] Add coverage in tests/test_predict.py for a mixed legacy/live session.

Scope

One production file (termstory/predict.py, ~15 lines) plus tests/test_predict.py.

sanjana2505006:
## Summary
- Closes #463
- `_load_sessions()` now drops `is_legacy` commands before scoring
- Sessions with no non-legacy commands are omitted (all-legacy fast path kept)
- Add regression test for mixed legacy/live session

## Test plan
- [x] `pytest tests/test_predict.py::TestPredictorEmptyHistory::test_mixed_legacy_live_session_excludes_legacy_commands -q`
- [x] `pytest tests/test_predict.py -q`

### DIFF
--- a/termstory/predict.py
+++ b/termstory/predict.py
@@ -134,7 +134,11 @@ class Predictor:
                 if cmd_rows and all(bool(r[1]) for r in cmd_rows):
                     continue
 
-                cmds = [r[0] for r in cmd_rows if not _is_noise_command(r[0])]
+                non_legacy_cmds = [r[0] for r in cmd_rows if not bool(r[1])]
+                if not non_legacy_cmds:
+                    continue
+
+                cmds = [cmd for cmd in non_legacy_cmds if not _is_noise_command(cmd)]
                 project_name = p_name if p_name and p_name not in ("General / No Project", "") else "Other"
 
                 sessions.append({
--- a/tests/test_predict.py
+++ b/tests/test_predict.py
@@ -227,6 +227,34 @@ class TestPredictorEmptyHistory:
         finally:
             os.unlink(db_path)
 
+    def test_mixed_legacy_live_session_excludes_legacy_commands(self):
+        """#463: legacy commands in a mixed session must not feed prediction signals."""
+        now = datetime.now()
+        db_path = _make_db([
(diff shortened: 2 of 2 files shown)


bitflicker64 (approved):
Summary: `_load_sessions()` now drops `is_legacy` rows before noise filtering and skips sessions that have no live commands left. That matches #463.

`termstory/predict.py` only changes the load loop; `tests/test_predict.py` adds `test_mixed_legacy_live_session_excludes_legacy_commands`. CI is green on lint and Python 3.9-3.12. No unrelated files.

